PR #1, opened 2018-07-26
tweaks on the README

meleu:
I did one more reading in the README and that's the result.

@leiradel feel free to edit my PR changing anything to be as you prefer ;)

Cheers.


leiradel:
I've added some of the chages as part of the latest release

PR #6, opened 2018-10-01
Fix C98 warnings

RobLoach:
See https://github.com/libretro/RetroArch/commit/072848445c5bd1298be693e2cf4baecb83444587#commitcomment-30710083

### DIFF
--- a/include/rcheevos.h
+++ b/include/rcheevos.h
@@ -100,7 +100,7 @@ enum {
   RC_OPERAND_8_BITS,
   RC_OPERAND_16_BITS,
   RC_OPERAND_24_BITS,
-  RC_OPERAND_32_BITS,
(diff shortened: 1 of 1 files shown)


leiradel:
Thanks for the PR. Only one question: why `RC_FORMAT_FRAMES = 0`? Is it necessary to fix a C98 warning? I guess not, since there are other enums without the `=0`.

RobLoach:
Unsure, that was one of the changes that @twinaphex made. I can remove the `= 0` if it has no change on compilation.

RobLoach:
Fixed in the above commit. Thanks for the explanation!

leiradel:
@RobLoach can you change this PR to target the develop branch? Thanks

RobLoach:
@leiradel There you go :+1:

PR #10, opened 2018-10-29
support for rich presence

Jamiras:
I tried my best to make this look like the existing functions. Let me know if I did something that will cause this to not be compatible on the required platforms.

I've ported over all of the unit tests from the integration DLL, and added a couple more.

 **Changes** 

- I added a `rc_alloc_str` method to load a null-terminated string into the buffer.
- I changed `rc_format_value` to return the number of characters written to the buffer.
- I modified the `parse_` test helper functions to validate that the specified size was not exceeded when actually doing the parsing.

**Data structures**

* `rc_richpresence_t` is a container with two linked lists. 
  * The first is a list of display strings (`rc_richpresence_display_t`)
  * The second is a list of lookups (`rc_richpresence_lookup_t`). This list is only used for constructing the list of display strings. The display string parts have direct references back to the appropriate lookups to avoid having to resolve them each time the rich presence is evaluated.

* `rc_richpresence_lookup_t` is a container with a linked list of items (`rc_richpresence_lookup_item_t`) and a format.
  * `RC_FORMAT_LOOKUP` was added as an internal format value, so the other format values could be directly used.
    * The last item in the items list holds the default text to display if no other items match.
  * Non-lookup formats have an empty list (first = NULL)

* `rc_richpresence_display_t` is a container with a `rc_trigger_t` and a linked list of display segments.
  * The container is part of a list. Each container's trigger will be evaluated to determine which display string to use. If it's the last container in the list, the trigger is ignored and the container is automatically selected.
  * Each display segment (`rc_richpresence_display_part_t`) has a `rc_value_t` expression and a display format. 
    * The `rc_value_t` expression calculates the value to format.
    * If the display format is `RC_FORMAT_LOOKUP`, the associated lookup's list of items is traversed to find an appropriate label to display.
    * If the display format is `RC_FORMAT_STRING` (also newly added internal format value), the part's text is displayed.
    * Otherwise, `rc_format_value` is used to display the value.

### DIFF
--- a/src/rcheevos/trigger.c
+++ b/src/rcheevos/trigger.c
@@ -18,6 +18,8 @@ void rc_parse_trigger_internal(rc_trigger_t* self, int* ret, void* buffer, rc_sc
     if (*ret < 0) {
       return;
     }
+
+    self->requirement->next = 0;
   }
 
   while (*aux == 's' || *aux == 'S') {
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,12 @@ Module.symvers
 Mkfile.old
 dkms.conf
 
+# Visual Studio files
+Debug/
+Release/
+*.user
+.vs/
+
 # Repository specific
 test/test
 test/galaga_nes.h
--- a/include/rcheevos.h
+++ b/include/rcheevos.h
@@ -29,7 +29,8 @@ enum {
   RC_MISSING_CANCEL = -14,
   RC_MISSING_SUBMIT = -15,
   RC_MISSING_VALUE = -16,
-  RC_INVALID_LBOARD_FIELD = -17
+  RC_INVALID_LBOARD_FIELD = -17,
+  RC_MISSING_DISPLAY_STRING = -18
 };
 
 /*****************************************************************************\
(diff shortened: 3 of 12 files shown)


src/rcheevos/trigger.c
```diff
@@ -18,6 +18,8 @@ void rc_parse_trigger_internal(rc_trigger_t* self, int* ret, void* buffer, rc_sc
     if (*ret < 0) {
       return;
     }
+
+    self->requirement->next = 0;
```
Jamiras: This just ensures the core group's next pointer is initialized. As far as I could tell, nothing was trying to iterate the core group.

leiradel:
Can you change the `rc_alloc` call in `rc_alloc_str` to align to `char`? It'll save just a few bytes, I know, it's just my OCD.

leiradel:
The `rc_alloc` call in `rc_alloc_str` is passing 0 as the scratch structure. This should crash when making a dry run to compute the amount of memory necessary to load the patch data.

In this case, since we don't know beforehand how many bytes will be needed for the biggest string, we should explicitly check the `pointer` argument, and just return `length + 1` if it's 0.

Jamiras:
I can definitely chance the alignment in `rc_alloc_str`. 

Passing 0 was a design choice as the length of the string was unknown. It will result in returning a null pointer for the allocated string, despite reserving an appropriate amount of space. Nothing is currently reading this string in the sizing logic, so I felt this was safe. 

I could use `min(length + 1, sizeof(rc_scratch_t))` if you'd prefer. I'm concerned that writing a string into scratch isn't safe as another operation may write a pointer over the string destroying the null terminator.

PR #11, opened 2018-11-28
Add "prior" memory type

Jamiras:
Supports https://github.com/RetroAchievements/RAIntegration/issues/176

Works like delta, but reports the last different value, instead of the value from the previous frame.

For example:

| Frame | Value | Delta | Prior |
| ----- | ----- | ----- | ----- |
|    1 |   25 |    0 |    0 |
|    2 |   25 |   25 |    0 |
|    3 |   30 |   25 |   25 |
|    4 |   35 |   30 |   30 |
|    5 |   35 |   35 |   30 |   
|    6 |   35 |   35 |   30 |
|    7 |   40 |   35 |   35 |

Represented by placing a 'P' or 'p' before the '0x' in a memory reference:
```
p0x1234
```
Can be combined with any existing logic:
```
p0x1234>0                // prior value is greater than 0
R:p0x1234!=0             // reset if prior value is not 0
P:p0x1234=10             // pause if prior value is 10
R:p0x1234>p0x1234         // reset if value is greater than prior value
R:d0x1234!=p0x1234       // reset if delta value differs from prior value
```

Note that `prior` will equal `value` only if the `value` has been 0 since resetting. Once `value` changes, `prior` will always be one "value" behind.

### DIFF
--- a/include/rcheevos.h
+++ b/include/rcheevos.h
@@ -109,7 +109,8 @@ enum {
   RC_OPERAND_DELTA,   /* The value last known at this address. */
   RC_OPERAND_CONST,   /* A 32-bit unsigned integer. */
   RC_OPERAND_FP,      /* A floating point value. */
-  RC_OPERAND_LUA      /* A Lua function that provides the value. */
+  RC_OPERAND_LUA,     /* A Lua function that provides the value. */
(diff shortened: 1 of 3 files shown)


meleu:
uh... same request as that one on the other PR: could you provide a RAIntegration build with this feature for testing?

Thanks in advance.

PR #12, opened 2018-11-28
add AndNext condition flag

Jamiras:
Supports https://github.com/RetroAchievements/RAIntegration/issues/6 and https://github.com/RetroAchievements/RAIntegration/issues/3

New condition flag `N:` indicates the condition _and_ the following condition must both be true at the same time. Multiple conditions may be joined to create more complicated dependencies. The final condition may use the ResetIf or PauseIf flag.

`N:0xH0001=20_N:0xH0002=20_0xH0003=20.3.`
indicates that $0001, $0002, and $0003 must _all_ be 20 in the same frame for three frames.

`0xH0000=0.100._N:0xH0001=13_R:0xH0002=86`
counts 100 frames of $0000 being 0, but restarts the counter if $0001 is 13 _and_ $0002 is 86.

`N:0xH0001=18_d0xH0001!=18.5.`
is only true if $0001 changes from any value to 18 five times.


meleu:
@Jamiras could you please provide a RAIntegration binary for testing when submitting a PR like this?

I'm sure that there are some enthusiastic guys that would love to make some tests and provide feedback (example: @kdecks and @Keltron3030)

Jamiras:
Not at this time. No changes have been made to the dll as of yet. Because the DLL is using rcheevos as a submodule, it's really hard reference the changes in my forked repository to make the DLL changes - that's one of the drawbacks of moving everything to rcheevos. I'm relying on the unit tests for functional testing.

It's probably "proper" to do all the extra work to ensure everything is correct before merging the rcheevos portion. I'll think about this some more - I'll probably have to make one-off builds without proper submodule references.

PR #13, opened 2018-12-12
Support floating point value starting with decimal point.

Jamiras:
Noticed while playing the AotW game: Zelda - Link to the Past (GBC). The Rich Presence is defined as follows:
```
@Hearts(0xha34c*.125) Hearts...
```
The ".125" was confusing the parser. Because it didn't start with a number it fell into the default case for an address, which didn't recognize the period and ignored it.

### DIFF
--- a/src/rcheevos/operand.c
+++ b/src/rcheevos/operand.c
@@ -256,6 +256,7 @@ static int rc_parse_operand_term(rc_operand_t* self, const char** memaddr, lua_S
       }
 
       /* fall through */
+    case '.':
     case '+': case '-':
     case '1': case '2': case '3': case '4': case '5':
     case '6': case '7': case '8': case '9':
--- a/test/test.c
+++ b/test/test.c
@@ -13,7 +13,7 @@
 
 typedef struct {
   unsigned char* ram;
-  int size;
+  unsigned size;
 }
 memory_t;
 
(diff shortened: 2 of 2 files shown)


src/rcheevos/operand.c
```diff
@@ -256,6 +256,7 @@ static int rc_parse_operand_term(rc_operand_t* self, const char** memaddr, lua_S
       }
 
       /* fall through */
+    case '.':
```
Jamiras: this is the actual fix

test/test.c
```diff
@@ -13,7 +13,7 @@
 
 typedef struct {
   unsigned char* ram;
-  int size;
+  unsigned size;
```
Jamiras: fixes warning about signed/unsigned comparison on line 21

test/test.c
```diff
@@ -608,6 +608,8 @@ static rc_condset_t* trigger_get_set(rc_trigger_t* trigger, int ndx) {
 }
 
 static void test_trigger(void) {
+  char buffer[2048];
```
Jamiras: fixes warning about possibly exceeding stack size. each test was apparently allocating its own 2K buffer on the stack

test/test.c
```diff
@@ -1532,6 +1513,7 @@ static void test_term(void) {
     parse_comp_term_fp("0xH1234*1", RC_OPERAND_8_BITS, 0x1234U, 1.0);
     parse_comp_term_fp("0xH1234*3", RC_OPERAND_8_BITS, 0x1234U, 3.0);
     parse_comp_term_fp("0xH1234*0.5", RC_OPERAND_8_BITS, 0x1234U, 0.5);
+    parse_comp_term_fp("0xH1234*.5", RC_OPERAND_8_BITS, 0x1234U, 0.5);
```
Jamiras: test for fix

PR #14, opened 2018-12-24
update delta values for paused conditions

Jamiras:
Implements https://github.com/RetroAchievements/RAIntegration/issues/144

* Renamed `RC_OPERAND_xxx` to `RC_MEMSIZE_xxx` for memory size enum
* Extracted unnamed struct from `rc_operand_t` as `rc_memref_value_t`. `rc_operand_t` now has a pointer to a `rc_memref_value_t`. This allows multiple conditions to reference the same `rc_memref_value_t`, and the console memory only has to be polled once per address/size combination.
  * As a result, the `sizeof (rc_operand_t)` has shrunk 8 bytes, but `sizeof(rc_memref_value_t)` is 24  bytes. Each `rc_condition_t` has two `rc_operand_t`s, and I moved a pointer so the structs are 16-byte aligned so `sizeof(rc_condition_t)` is down 24 bytes. For a condition that compares a memory value to a constant, there's no net change in memory consumption. If both sides are memory references, more memory will be used, but that will be regained if any addresses are duplicated.
  * Various trigger sizes (before and after):
    * `1=1`: 96 before, 72 after (no memory references)
    * `0xh1234=1`: 96 before, 96 after (one memory reference)
    * `0xh1234!=d0xh1234`: 96 before, 96 after (one shared memory reference)
    * `0xh1234!=0xh1235`: 96 before, 120 after (two memory references)
    * `0xH0013f9=3_R:0xH0013f9!=3_0xH0013ef=1_R:0xH0013ef!=1S0xH0013bf=50S0xH0013bf=49`: 
      488 before, 416 after (three conditions, three memory references)
    * `0xR00625e=1_0xP006260=1_0xO006262=1_0xN006262=1_0xR006263=1_0xQ006263=1_0xO006264=1_0xM006266=1_0xO006267=1_0xR00626b=1_0xN00626b=1_0xO00626f=1_0xQ006270=1_0xS006272=1_0xR006272=1_0xQ006272=1_0xT006273=1_0xS006273=1_0xR006273=1_0xO006273=1_0xR006274=1_0xQ006274=1_0xP006274=1_0xN006274=1_0xM006274=1_0xR006275=1_0xQ006275=1_0xR006276=1_0xQ006276=1_0xP006276=1_0xO006276=1_0xN006276=1_0x 000042!=0Sd0xR00625e=0Sd0xP006260=0Sd0xO006262=0Sd0xN006262=0Sd0xR006263=0Sd0xQ006263=0Sd0xO006264=0Sd0xM006266=0Sd0xO006267=0Sd0xR00626b=0Sd0xN00626b=0Sd0xO00626f=0Sd0xQ006270=0Sd0xS006272=0Sd0xR006272=0Sd0xQ006272=0Sd0xT006273=0Sd0xS006273=0Sd0xR006273=0Sd0xO006273=0Sd0xR006274=0Sd0xQ006274=0Sd0xP006274=0Sd0xN006274=0Sd0xM006274=0Sd0xR006275=0Sd0xQ006275=0Sd0xR006276=0Sd0xQ006276=0Sd0xP006276=0Sd0xO006276=0Sd0xN006276=0`: 
      5216 before, 4448 after (65 conditions, 33 memory references)
* Each call to `rc_test_trigger`, `rc_evaluate_lboard`, `rc_evaluate_value`, and `rc_evaluate_richpresence` will traverse the `rc_memref_value_t` chain to evaluate the memory references before doing any conditional logic. this is the primary functional change for this change.
* `rc_leaderboard_t` contains three `rc_trigger_t`s and a `rc_value_t`. They share a single `rc_memref_value_t` chain.
* Added `rc_parse_state` container to hold parameters passed to parsing functions. The internal function signatures were getting lengthy.

_NOTE_: The first two changes modify the public contract for the data types. At this point in time, the only thing using the raw data objects is the RAIntegration.dll, and it will be updated accordingly. The RetroArch integration is only storing the records and not examining them.

I have made appropriate changes to the DLL to leverage these changes and have been running with them for over a week (will open a PR once this is merged). I believe the changes to be correct and functional. Please review carefully.


include/rcheevos.h
```diff
@@ -171,9 +192,6 @@ enum {
 typedef struct rc_condition_t rc_condition_t;
 
 struct rc_condition_t {
-  /* The next condition in the chain. */
-  rc_condition_t* next;
```
Jamiras: putting this before the `rc_operand_t` structs allocated an extra buffer between `next` and `operand1` so `operand1` would be 8-byte aligned.

src/rcheevos/alloc.c
```diff
@@ -10,22 +11,43 @@ void* rc_alloc(void* pointer, int* offset, int size, int alignment, rc_scratch_t
   if (pointer != 0) {
     ptr = (void*)((char*)pointer + *offset);
   }
+  else if (scratch != 0) {
+    ptr = &scratch->obj;
+  }
```
Jamiras: `rc_alloc_str` calls `rc_alloc` with a NULL `scratch`, so we can't transpose the subobject to return null. Had to add an explicit check

src/rcheevos/alloc.c
```diff
@@ -10,22 +11,43 @@ void* rc_alloc(void* pointer, int* offset, int size, int alignment, rc_scratch_t
...
   return ptr;
 }
+
+void rc_init_parse_state(rc_parse_state_t* parse, void* buffer, lua_State* L, int funcs_ndx)
+{
+  parse->offset = 0;
+  parse->L = L;
+  parse->funcs_ndx = funcs_ndx;
+  parse->buffer = buffer;
+  parse->scratch.memref = parse->scratch.memref_buffer;
```
Jamiras: scratch provides 16 memrefs, if more are needed, we'll allocate then.

src/rcheevos/alloc.c
```diff
@@ -10,22 +11,43 @@ void* rc_alloc(void* pointer, int* offset, int size, int alignment, rc_scratch_t
...
+  parse->scratch.memref_size = sizeof(parse->scratch.memref_buffer) / sizeof(parse->scratch.memref_buffer[0]);
+  parse->scratch.memref_count = 0;
+  parse->first_memref = 0;
+}
+
+void rc_destroy_parse_state(rc_parse_state_t* parse)
+{
+  if (parse->scratch.memref != parse->scratch.memref_buffer)
+    free(parse->scratch.memref);
```
Jamiras: if we allocated, clean up

src/rcheevos/memref.c
```diff
@@ -0,0 +1,204 @@
...
+
+void rc_update_memref_values(rc_memref_value_t* memref, rc_peek_t peek, void* ud) {
+  while (memref) {
+    memref->previous = memref->value;
+    memref->value = rc_memref_get_value(&memref->memref, peek, ud);
+    if (memref->value != memref->previous)
+      memref->prior = memref->previous;
+
+    memref = memref->next;
+  }
```
Jamiras: memref processing loop traverses chain and updates the three data values appropriately

src/rcheevos/memref.c
```diff
@@ -0,0 +1,204 @@
...
+}
+
+void rc_init_parse_state_memrefs(rc_parse_state_t* parse, rc_memref_value_t** memrefs)
+{
+  parse->first_memref = memrefs;
+  *memrefs = 0;
+}
```
Jamiras: use separate function to initialize memref chain so chain is associated to parser, not to individual triggers. allows leaderboard to share memref chain across start/submit/cancel triggers.

src/rcheevos/operand.c
```diff
@@ -264,18 +269,24 @@ static int rc_parse_operand_term(rc_operand_t* self, const char** memaddr, lua_S
...
       self->fp_value = strtod(aux, &end);
 
       if (end == aux) {
         return RC_INVALID_FP_OPERAND;
       }
 
+      if (floor(self->fp_value) == self->fp_value) {
+        self->type = RC_OPERAND_CONST;
+        self->value = (unsigned)floor(self->fp_value);
+      }
```
Jamiras: minor optimization: if the multiplier is a whole number, we can do integer multiplication instead of converting to a float, multiplying, and converting back.

src/rcheevos/operand.c
```diff
@@ -368,113 +374,15 @@ unsigned rc_evaluate_operand(rc_operand_t* self, rc_peek_t peek, void* ud, lua_S
...
-                  + ((value >> 12) & 0x0f) *     1000
-                  + ((value >>  8) & 0x0f) *      100
-                  + ((value >>  4) & 0x0f) *       10
-                  + ((value >>  0) & 0x0f) *        1;
-          }
-
-          break;
-      }
+      value = self->memref->value;
+      break;
```
Jamiras: When processing the conditions, we just have to look at the already updated memory references.

src/rcheevos/operand.c
```diff
@@ -368,113 +374,15 @@ unsigned rc_evaluate_operand(rc_operand_t* self, rc_peek_t peek, void* ud, lua_S
...
-      if (self->type == RC_OPERAND_DELTA) {
-        unsigned previous = self->previous;
-        self->previous = value;
-        value = previous;
-      } else if (self->type == RC_OPERAND_PRIOR) {
-        if (self->previous != value) {
-          self->prior = self->previous;
-          self->previous = value;
-        }
-        value = self->prior;
```
Jamiras: delta and prior are calculated unconditionally for all memory references so they're available for all references. it also eliminates a couple of branches in the processing code.

rzumer:
Who has access to this repository, only @leiradel ? It is a blocker for some of my work so I would like it merged whenever possible.

meleu:
uh... I think my last PR (#18, changing some lines `operand.c`) made this branch have conflicts.
@Jamiras can you solve it?

PR #21, opened 2019-07-06
Add flag for tracking HitCount

Jamiras:
When a condition is flagged with the "Measured" attribute (`M:`), it won't affect the evaluation of the trigger, but will provides metadata to the runtime. For example:

image

This achievement is measuring a HitCount that is currently at 191/250. This information could be used to show progress in the overlay (https://github.com/RetroAchievements/RAIntegration/issues/4):

image

It is expected that only one Measured flag be present per achievement, and that the Measured condition have a target HitCount. If more than one exists, only the first will be tracked. If no target HitCount is present, progress cannot be tracked.

Additionally, this marker can be used to tally something happening for leaderboards (https://github.com/RetroAchievements/RAIntegration/issues/13). Putting the following into the VALUE field of the leaderboard:
```
M:0xH1234!=d0xH1234
```
Would submit the number of times the byte at $1234 changed while the leaderboard was active. 

A "`M:`" prefix or a "`C:`" (AddHits) prefix indicates that a condition should be parsed instead of a value. The "`M:`" prefix is required, but may be preceeded by any number of "`C:`" prefixes. The old value format is still fully supported. 

Counters will always start at 0, and there's currently no way to modify the result (i.e. add 2 or count down from some arbitrary value).


Keltron3030:
Really cool 👍

PR #24, opened 2019-07-16
add SIGNED format type

Jamiras:
`rc_evaluate_value` returns an unsigned int, and formatting for `SCORE`, `VALUE`, and `OTHER` types also treats the value as unsigned. There are cases (such as [this leaderboard](https://retroachievements.org/leaderboardinfo.php?i=2639)) where negative values are desired. With the rcheevos implementation, a value above `MAX_INT` was submitted and truncated to `MAX_INT`.

This solution adds a `SIGNED` format, which will cast the unsigned value back to a signed value and use a signed formatter to convert the value to a string. This will address the on-screen display, but further changes will be required in the DLL to submit a negative value to the server.

Additionally, the leaderboard editor page will need to be updated to allow specifying `SIGNED` as the format for the leaderboard.

This also opens up the `SIGNED` format for use in rich presence, with no further changes required in the DLL or website:
```
Format:Money
FormatType=SIGNED

Display:$@Money(0xx2045)
```

Note: this assumes 32-bit signed values. If reading less than 32-bits from memory, the sign bit would have to be extended:
```
Display:$@Money(0x 2045_hffff0000*0xt2046)
```
`t` is the uppermost bit of the 16-bit value, and the multiplication sets all the bits in the upper 16-bits if the test bit was set.

### DIFF
--- a/include/rcheevos.h
+++ b/include/rcheevos.h
@@ -338,7 +338,8 @@ enum {
   RC_FORMAT_CENTISECS,
   RC_FORMAT_SCORE,
   RC_FORMAT_VALUE,
-  RC_FORMAT_OTHER
+  RC_FORMAT_OTHER,
+  RC_FORMAT_SIGNED
 };
 
 int rc_parse_format(const char* format_str);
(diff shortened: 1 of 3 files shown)


Jamiras:
Changed my mind on this. The DB and server API both only handle signed ints currently, so going to change the `rc_evaluate_value` to return a signed int instead.

PR #69, opened 2020-05-23
increase message sizes

meleu:
In order to silence warnings when compiling RetroArch on Linux+GCC 8.3.0.

See: https://github.com/libretro/RetroArch/commit/268bf61da65d23fe1413e87207d43ea1acd7b64c#commitcomment-39377709


Jamiras:
I've provided an alternate solution in #70. It addresses the issue by changing the format strings instead of the buffer sizes.

meleu:
Closing this so we can find a cleaner solution.

PR #70, opened 2020-05-25
fix format-truncation warnings

Jamiras:
alternate approach to address issue reported in #69


test/Makefile
```diff
@@ -28,7 +28,7 @@ OBJ=$(RC_SRC)/trigger.o $(RC_SRC)/condset.o $(RC_SRC)/condition.o $(RC_SRC)/oper
 CFLAGS=-Wall -O0 -g -std=c89 -ansi -Wno-long-long -I../include -I$(RC_SRC)
 
 # more strict validation for source files to eliminate warnings/errors in RetroArch compile
-SRC_CFLAGS=-pedantic -Wsign-compare
+SRC_CFLAGS=-pedantic -Wsign-compare -std=gnu99 -D_GNU_SOURCE -Wall
```
Jamiras, line 31: I'm a bit concerned about this change. I had to copy these settings from RetroArch to get the warning to appear, even after switching to GCC 8.3. This changes the compilation target from c89 to gnu99 (gnu89 does not generate the errors). If strict c89 compliance is required, this may allow some things to slip through while enforcing the newly required warnings.

PR #72, opened 2020-05-26
include parent folder in arcade hash for specific folder names

Jamiras:
FBNeo supports several consoles in addition to arcade via subsystems. Because the arcade hash is just the md5 of the filename, this can lead to conflicts between actual arcade roms and console roms. To address this, I've been asked to include the parent folder as part of the hash when the parent folder is one of the [known subsystem keys](https://github.com/libretro/FBNeo/blob/master/src/burner/libretro/README.md#emulating-consoles).

For example, `columns.zip` could be the arcade game or the megadrive game. If present in the `megadriv` folder, the hash would be calculated from `megadriv_columns`, which would differentiate it from `columns`.

Note that FBNeo subsystem games must still be launched via the `Arcade > FBNeo` core in RALibretro. This may be confusing to users, but we don't really expect many to use FBNeo as their primary core for the consoles that FBNeo supports.

While this solution is primarily to avoid loading an arcade set for a non-arcade title, it's entirely possible that the non-arcade hash could be added. It would have to be done manually as the unknown game dialog would only show arcade titles.

### DIFF
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -505,6 +505,7 @@ static int rc_hash_arcade(char hash[33], const char* path)
             memcmp(folder, "fds", 3) == 0 ||
             memcmp(folder, "sms", 3) == 0 ||
             memcmp(folder, "msx", 3) == 0 ||
+            memcmp(folder, "ngp", 3) == 0 ||
             memcmp(folder, "pce", 3) == 0 ||
             memcmp(folder, "sgx", 3) == 0)
           include_folder = 1;
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -475,9 +475,73 @@ static int rc_hash_3do(char hash[33], const char* path)
 static int rc_hash_arcade(char hash[33], const char* path)
 {
   /* arcade hash is just the hash of the filename (no extension) - the cores are pretty stringent about having the right ROM data */
-  const char* ptr = rc_path_get_filename(path);
-  const char* ext = rc_path_get_extension(ptr);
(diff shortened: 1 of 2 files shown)


Sanaki:
You'll want to add ngp as well, which was only added a couple days ago. It hasn't been added to that readme yet.

Sanaki:
I attempted to test the changes by replacing hash.c in RetroArch's source with this one and rebuilding, but found no change in the generated hashes from doing so. Either I'm missing something or something's going awry.

Jamiras:
retroarch doesn't currently call into hash.c. you have to test in RALibRetro.

Sanaki:
Found time to test it properly using RAHasher. As I had experienced with CLI loading the console games in retroarch (an unlikely scenario outside testing), the one caveat to this is that hashing the file from within the directory won't see the subdirectory.
```
RAHasher 27 coleco_centiped.7z
ecd2be6e18a47dac27aded4390a2b23f
RAHasher 27 centiped.7z
3e30300051e145abaf2164a53f107bb7
RAHasher 27 coleco/centiped.7z
ecd2be6e18a47dac27aded4390a2b23f
cd coleco
RAHasher 27 centiped.7z
3e30300051e145abaf2164a53f107bb7
```
In theory, this should -only- affect RAHasher, since RALibretro and RetroArch should be receiving the full path at all times.

Jamiras:
The related ticket has been closed in a way that doesn't seem to affect these changes. Do you see any reason to not merge this?

Sanaki:
I don't. Looks fine to me.

PR #87, opened 2020-07-21
use default case statement for minor performance gain

Jamiras:
`rc_evaluate_operand` is called a lot when evaluating achievements. This eliminates the need to explicitly check three operand types and just assumes the anything not otherwise handled is a memory reference.

Provided a 1-2% performance gain when simulating 10 minutes of frame evaluation against a handful of games on a release build.

### DIFF
--- a/src/rcheevos/operand.c
+++ b/src/rcheevos/operand.c
@@ -312,7 +312,7 @@ unsigned rc_evaluate_operand(rc_operand_t* self, rc_eval_state_t* eval_state) {
   rc_luapeek_t luapeek;
 #endif /* RC_DISABLE_LUA */
 
-  unsigned value = 0;
+  unsigned value;
 
   /* step 1: read memory */
   switch (self->type) {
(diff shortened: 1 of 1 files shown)


leiradel:
Nice and intriguing!

meleu:
I became intrigued about that too, and after some googling it seems to me that the compiler optimization is doing its job

I've found an interesting info that can shed some light on this fact.

[Efficient C Tip #12 – Be wary of switch statements](https://embeddedgurus.com/stack-overflow/2010/04/efficient-c-tip-12-be-wary-of-switch-statements/) (bold highlights are mine)

> ## Order of Execution
> 
> For many years I wrote switch statements under the assumption that the case values would be evaluated from top to bottom. That is, if the compiler chose to implement the switch statement as an if-else-if chain, then it would first test the first case, then the second case and so on down to the default case at the bottom of my source code. Well it turns out that my assumption was completely wrong. The compiler is under no such obligation, and indeed will often evaluate the values bottom to top. Furthermore, **the compiler will often evaluate the default value first. For example consider a defaulted switch statement with contiguous case values in the range zero to ten. If the index variable is an unsigned int, then there are at least 65525 possible values handled by the default case, and so it makes sense to eliminate them first**. Now if you know that the index variable can only possibly take on the values zero to ten, then you can of course eliminate the default statement – and then get excoriated by the coding standards / MISRA folks.

leiradel:
That's not a good article on how compilers generate code for `switch` statement. The author is right when he says that the compiler is free to generate the code in any way it wants, but there are too many assumptions there. What the compiler does depends on the compiler itself, its version, the target architecture, and the phase of the moon.

I've seen `if`-`else` chains (for small number of cases), jump tables (for contiguous values), and binary searches (for too many cases with uncontiguous values), and it's very difficult to predict which one will be used. I've never seen bottom-up evaluation.

I'd think that in this case the compiler would generate a jump table since the case values are near each other, but maybe it wasn't. Maybe with the reduced number of cases, a couple of comparisons became faster than a jump table. The only way to be sure is to look at the generated code before and after the change.

PR #91, opened 2020-08-20
Add hash validation for Dreamcast GDI files

miagui:
NEW COMMENT
GDI validation is done by hashing a 256 bytes header at track 3 and the boot executable specified on header (usually it is 1ST_READ.BIN). The executable is found at the last track.
Source for information: https://mc.pp.se/dc/ip.bin.html

OLD COMMENT
Dreamcast hashing is almost the same as for Sega CD and Saturn, except that it hashes 256 bytes.

First, i've created a function for reading .GDI files and getting a track, as their format are a bit different from .CUE. Then it reads the first sector from the first track (it should always have unique meta information in the first 16 sectors, see: https://mc.pp.se/dc/ip0000.bin.html) and make a hash with it's contents.

### DIFF
--- a/include/rhash.h
+++ b/include/rhash.h
@@ -105,9 +105,8 @@ extern "C" {
   /* closes the track handle */
   typedef void (*rc_hash_cdreader_close_track_handler)(void* track_handle);
 
+  /* get number of tracks */
   typedef int (*rc_hash_cdreader_num_tracks_handler)(const char* path);
-
-  typedef uint32_t(*rc_hash_cdreader_get_lba_handler)(void* track_handle);
 
   struct rc_hash_cdreader
   {
@@ -115,7 +114,6 @@ extern "C" {
     rc_hash_cdreader_read_sector_handler          read_sector;
     rc_hash_cdreader_close_track_handler          close_track;
     rc_hash_cdreader_num_tracks_handler           num_tracks;
-    rc_hash_cdreader_get_lba_handler              get_lba;
   };
 
   void rc_hash_init_default_cdreader();
--- a/include/rhash.h
+++ b/include/rhash.h
@@ -105,14 +105,17 @@ extern "C" {
   /* closes the track handle */
   typedef void (*rc_hash_cdreader_close_track_handler)(void* track_handle);
 
+  typedef int (*rc_hash_cdreader_num_tracks_track_handler)(const char* path);
+
   typedef uint32_t(*rc_hash_cdreader_get_lba_handler)(void* track_handle);
 
   struct rc_hash_cdreader
   {
-    rc_hash_cdreader_open_track_handler      open_track;
-    rc_hash_cdreader_read_sector_handler     read_sector;
-    rc_hash_cdreader_close_track_handler     close_track;
-    rc_hash_cdreader_get_lba_handler         get_lba;
+    rc_hash_cdreader_open_track_handler           open_track;
+    rc_hash_cdreader_read_sector_handler          read_sector;
+    rc_hash_cdreader_close_track_handler          close_track;
+    rc_hash_cdreader_num_tracks_track_handler     num_tracks;
+    rc_hash_cdreader_get_lba_handler              get_lba;
   };
 
   void rc_hash_init_default_cdreader();
--- a/include/rhash.h
+++ b/include/rhash.h
@@ -105,7 +105,7 @@ extern "C" {
   /* closes the track handle */
   typedef void (*rc_hash_cdreader_close_track_handler)(void* track_handle);
 
-  typedef int (*rc_hash_cdreader_get_lba_handler)(void* track_handle);
+  typedef uint32_t(*rc_hash_cdreader_get_lba_handler)(void* track_handle);
 
   struct rc_hash_cdreader
   {
--- a/include/rhash.h
+++ b/include/rhash.h
@@ -105,11 +105,14 @@ extern "C" {
   /* closes the track handle */
   typedef void (*rc_hash_cdreader_close_track_handler)(void* track_handle);
 
+  typedef int (*rc_hash_cdreader_get_lba_handler)(void* track_handle);
+
   struct rc_hash_cdreader
   {
     rc_hash_cdreader_open_track_handler      open_track;
     rc_hash_cdreader_read_sector_handler     read_sector;
     rc_hash_cdreader_close_track_handler     close_track;
+    rc_hash_cdreader_get_lba_handler         get_lba;
   };
 
   void rc_hash_init_default_cdreader();
--- a/include/rhash.h
+++ b/include/rhash.h
@@ -105,7 +105,7 @@ extern "C" {
   /* closes the track handle */
   typedef void (*rc_hash_cdreader_close_track_handler)(void* track_handle);
 
-  typedef int (*rc_hash_cdreader_num_tracks_track_handler)(const char* path);
+  typedef int (*rc_hash_cdreader_num_tracks_handler)(const char* path);
 
   typedef uint32_t(*rc_hash_cdreader_get_lba_handler)(void* track_handle);
 
@@ -114,7 +114,7 @@ extern "C" {
     rc_hash_cdreader_open_track_handler           open_track;
     rc_hash_cdreader_read_sector_handler          read_sector;
     rc_hash_cdreader_close_track_handler          close_track;
-    rc_hash_cdreader_num_tracks_track_handler     num_tracks;
+    rc_hash_cdreader_num_tracks_handler           num_tracks;
     rc_hash_cdreader_get_lba_handler              get_lba;
   };
 
--- a/src/rhash/cdreader.c
+++ b/src/rhash/cdreader.c
@@ -546,13 +546,16 @@ static void* cdreader_open_gdi_track(const char* path, uint32_t track)
     while (*ptr != '\n' && ptr < end)
     {
       ++ptr;
+      if (*ptr == '\n')
+      {
+        ++ptr;
+        break;
+      }
     }
 
     /* begins looping inside content*/
     while (ptr < end)
     {
-      ++ptr;
-
       ptr2 = ptr;
 
       while ((*ptr2 >= 97 && *ptr2 <= 122) || (*ptr2 >= 65 && *ptr2 <= 90) ||
@@ -576,8 +579,10 @@ static void* cdreader_open_gdi_track(const char* path, uint32_t track)
       }
 
       /* skip newlines */
-      while (*ptr == '\r')
-        ++ptr;
+      while ((*ptr == '\n' && ptr < end) || (*ptr == '\r' && ptr < end))
+      {
+          ++ptr;
+      }
     }
     if (done)
       break;
--- a/src/rhash/cdreader.c
+++ b/src/rhash/cdreader.c
@@ -625,17 +625,20 @@ static void* cdreader_open_gdi_track(const char* path, uint32_t track)
   char mode[16] = "MODE1/";
   char sector_size[16];
   char file[256];
+  size_t track_size;
+  int track_type;
   char* bin_path = "";
   int current_track = 0;
   char* ptr, * ptr2, * end;
 
   int offset = 0;
   int lba = 0;
-  int highest_lba = 0;
-  int highest_track = 0;
+  int largest_track = 0;
+  size_t largest_track_size = 0;
 
   int found = 0;
   size_t num_read = 0;
+  size_t file_offset = 0;
   struct cdrom_t* cdrom = NULL;
 
   file_handle = rc_file_open(path);
@@ -657,7 +660,7 @@ static void* cdreader_open_gdi_track(const char* path, uint32_t track)
 
     ptr = buffer;
 
-    /* removes first line, as it always have a track counter */
+    /* skip until next newline */
     while (*ptr != '\n' && ptr < end)
     {
       ++ptr;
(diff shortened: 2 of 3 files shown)


Sanaki:
Any chance you could make this applicable to gdi-based chd as well? Lacking chd support for Dreamcast would be... problematic. Especially considering that every single gdi image ends up being padded to >1GB without chd involved.

miagui:
Sure, i didn't know that chd was important to consider, but i'll try to do it.

Sanaki:
While RALibretro doesn't support chd yet, rcheevos is used in RetroArch as well.

Jamiras:
CHD handling is managed outside of rcheevos by the rc_hash_cdreader hooks. If no hooks are installed, the default cdreader functions are used, which only support uncompressed files (bin/cue, iso). We don't want rcheevos to be explicitly dependent on any other libraries.

If I understand the formats correctly (I haven't even started looking at dreamcast hashing), a gdi file is similar to a cue file and just tells the code where individual tracks start in the bin file. As such, this seems like an appropriate thing to have in rcheevos, but really makes me question this statement:

> Any chance you could make this applicable to gdi-based chd as well?

How would a CHD created from a GDI be any different than a CHD created from a CUE? The track offsets will be stored internally within the CHD and the source should be irrelevant.

Additionally:

> Dreamcast hashing is almost the same as for Sega CD and Saturn, except that it hashes 256 bytes.

I disagree with this statement. The document you've linked specifically states that that is just the header. The boot filename is present in the header, so we should attempt to hash that as well. That would be more consistent with the other CD file formats. By at least validating the code, we can ensure players aren't using games with baked in cheats.

I have several additional comments on the code itself before even trying to validate the functionality. Travis appears to be unhappy with it as well.

src/rhash/hash.c
```diff
@@ -868,6 +868,32 @@ static int rc_hash_pcfx_cd(char hash[33], const char* path)
...
+
+  track_handle = rc_cd_open_track(path, 1);
+  if (!track_handle)
+    return rc_hash_error("Could not open track");
+
+  // first sector from the first track should always have a IP0000.BIN structure that stores unique meta information.
+  // "The structure described below is repeated in the 16 first sectors of the first Mode-1 track on the disc".
+  // https://mc.pp.se/dc/ip0000.bin.html
+  rc_cd_read_sector(track_handle, 0, buffer, sizeof(buffer));
+
```
Jamiras, line 884: you should validate the header here to make sure it is a dreamcast disc.
```
if (memcmp(&buffer[0], "SEGA SEGAKATANA ", 16) != 0) {
    rc_close_track(track_handle);
    return rc_hash_error("Not a Dreamcast CD");
}
```

src/rhash/hash.c
```diff
@@ -868,6 +868,32 @@ static int rc_hash_pcfx_cd(char hash[33], const char* path)
...
+  if (verbose_message_callback)
+  {
+    char message[300];
+    snprintf(message, sizeof(message), "Meta information:\n%.256s", &buffer[0x00]);
```
Jamiras, line 888: This will spit out the entire header block (assuming it doesn't contain any null characters). It would probably be better to just log the "Name of the software" field (maybe also "Product number" and "Product version" fields)

src/rhash/hash.c
```diff
@@ -868,6 +868,32 @@ static int rc_hash_pcfx_cd(char hash[33], const char* path)
...
+    verbose_message_callback(message);
+  }
+
+  rc_cd_close_track(track_handle);
+
+  return rc_hash_buffer(hash, buffer, sizeof(buffer));
```
Jamiras, line 894: You should attempt to locate the boot file (as indicated by the "Boot filename" field) and include that in the hash.

src/rhash/cdreader.c
```diff
@@ -508,10 +508,129 @@ static void* cdreader_open_cue_track(const char* path, uint32_t track)
...
+
+        if (current_track == (int)track)
+        {
+          done = 1;
+          break;
+        }
+      }
+
+      // skip newlines
+      while (*ptr == '\r')
```
Jamiras, line 580: see previous comments about '\r' vs '\n' and reading past the end of the buffer

src/rhash/cdreader.c
```diff
@@ -508,10 +508,129 @@ static void* cdreader_open_cue_track(const char* path, uint32_t track)
...
+      while ((*ptr2 >= 97 && *ptr2 <= 122) || (*ptr2 >= 65 && *ptr2 <= 90) ||
+        (*ptr2 >= 48 && *ptr2 <= 57) || (*ptr2 == 32) || (*ptr2 == '.' || (*ptr2 == ' ')))
+      {
+        ++ptr2;
+      }
+
+      if (ptr2 - ptr != 0)
+      {
+        memcpy(track_data, ptr, ptr2 - ptr);  // operation to get track data
+        track_data[ptr2 - ptr] = '\0';
```
Jamiras, line 568: you can eliminate the copy (and the `track_data` buffer) by just setting `*ptr2 = '\0'`

src/rhash/cdreader.c
```diff
@@ -508,10 +508,129 @@ static void* cdreader_open_cue_track(const char* path, uint32_t track)
...
+      snprintf((char*)buffer, sizeof(buffer), "Failed to allocate %u bytes", (unsigned)sizeof(*cdrom));
+      rc_hash_error((const char*)buffer);
+      return NULL;
+    }
+
+    cdrom->first_sector_offset = offset;
+
+    bin_path = cdreader_get_bin_path(path, file);
+
+    if (cdreader_open_bin(cdrom, bin_path, mode)) // gives your cd rom
```
Jamiras, line 604: `cdreader_open_bin` is expecting a mode similar to `MODE2/2352`
The GDI format only specifies the sector size, not the full mode. But that's only used if the sector size can't be determined by looking at the contents of the bin. And there is a difference between `MODE1/2352` and `MODE2/2352` regarding how much header data is present for each sector. I'm not sure which is implied by the `2352` in the GDI file, but the presence of a sector header should allow the logic to determine the correct mode without being prompted.

src/rhash/cdreader.c
```diff
@@ -508,10 +508,129 @@ static void* cdreader_open_cue_track(const char* path, uint32_t track)
...
+        ++ptr;
+    }
+    if (done)
+      break;
+
+  } while (1);
+
+  uint8_t file_len = strlen(file);
+
+  if (file_len)
```
Jamiras, line 590: The `sscanf` will populate `file` for each line it processes, so this will be non-empty if any tracks exist. You should check `done` (maybe rename it to `found`) to ensure the `file` contents are actually significant.

miagui:
> I disagree with this statement. The document you've linked specifically states that that is just the header. The boot filename is present in the header, so we should attempt to hash that as well. That would be more consistent with the other CD file formats. By at least validating the code, we can ensure players aren't using games with baked in cheats.

That makes sense, i haven't considered that the boot file could be used to prevent players from cheating.

> I have several additional comments on the code itself before even trying to validate the functionality. Travis appears to be unhappy with it as well.

Thanks for reviewing, it seems that Travis has failed because i've used comments the wrong way, and "%s" didn't like specified size char* variables on sscanf.

src/rhash/cdreader.c
miagui: They seems to be MODE1, according to these docs i've found:

image

image

src/rhash/cdreader.c
miagui: Interesting, but is it worth it? Because i'd need to create another variable ptr3 to store progress from buffer reading.

Jamiras: Why would you need a third pointer? `ptr2` would maintain the progress, and you already use it in that way when you set `ptr = ptr2` on line 575
```
*ptr2++ = '\0'; /* null terminate and advance */

/* 1st try to get file if it have double quotes*/
sscanf(ptr, "%d %*s %*s %s \"%[^\"]\" %*s", &current_track, sector_size, file);

if (current_track == (int)track)
{
    ...

    if (strlen(file) == 0)
        sscanf(track_data, "%*d %*s %*s %*s %s %*s", file);

    ...
}

ptr = ptr2;
```

miagui: I undestand now, thanks for explaining.

miagui:
I've made so it hashes the boot executable, but it is a bit different process, so i did some changes.

It seems that the track describe the sector location of anything always being sector + LBA, so when searching for the directory record on `rc_cd_find_file_sector`, it gives the sector + LBA location, which leads the reader to nothing, so i needed to add a parameter on the function to subtract it by its LBA to get the real location on the file.

```
  static uint32_t rc_cd_find_file_sector(void* track_handle, const char* path, unsigned* size, unsigned sector_offset)
  {
  ...
    sector = buffer[156 + 2] | (buffer[156 + 3] << 8) | (buffer[156 + 4] << 16);
    sector -= sector_offset; 
  ...
  }
```

Because of this, i've added a LBA variable in `cdrom_t` struct, so it can be used later when doing these operations on `rc_hash_dreamcast`, but i've found no use for this in cuesheet files, as they don't state the LBA in there. Consequently, it may be unpractical to hash Dreamcast .cue files due to this.

include/rhash.h
```diff
@@ -105,11 +105,14 @@ extern "C" {
...
   typedef void (*rc_hash_cdreader_close_track_handler)(void* track_handle);
 
+  typedef uint32_t(*rc_hash_cdreader_get_lba_handler)(void* track_handle);
+
   struct rc_hash_cdreader
   {
     rc_hash_cdreader_open_track_handler      open_track;
     rc_hash_cdreader_read_sector_handler     read_sector;
     rc_hash_cdreader_close_track_handler     close_track;
+    rc_hash_cdreader_get_lba_handler         get_lba;
```
Jamiras, line 115: I really don't like this implementation. `get_lba` is retrieving supplemental data captured by open_track. I understand the need for it based on stepping through the code, but I think a cleaner solution is to encode the `lba` in `cdrom->first_sector_offset` as follows:
```
cdrom->first_sector_offset = -(lba * cdrom->sector_size);
```
This doesn't completely solve the problem though. `rc_cd_find_file_sector` would no longer be able to find the TOC data in sector 16, so you'd have to modify it to take a TOC sector parameter that would be set to 45016 for Dreamcast. This solution would keep the "CD" interface clean and abstract the gdi-specific behavior to the Dreamcast hashing code.

I also recommend using an overloaded method instead of changing all the calls. An extra parameter of 0 (or 16 in this case) has very little meaning to people reading the code. Hiding it in a helper function makes the code more readable.
```
static uint32_t rc_cd_find_file_sector_toc(void* track_handle, const char* path, unsigned* size, 
                                           unsigned toc_sector)
{
    .. existing function, modified to use toc_sector parameter
}

static uint32_t rc_cd_find_file_sector(void* track_handle, const char* path, unsigned* size)
{
    /* the cd information is usually 16 frames in */
    return rc_cd_find_file_sector_toc(track_handle, path, size, 16);
}
``` 
The resulting sector would then automatically be adjusted to the `last_track_handle`'s `first_sector_offset`, finding the binary without having to remember to make another adjustment based on the LBA of the last track.

src/rhash/cdreader.c
```diff
@@ -508,10 +513,155 @@ static void* cdreader_open_cue_track(const char* path, uint32_t track)
...
+      }
+
+      /* skip newlines */
+      while ((*ptr == '\n' && ptr < end) || (*ptr == '\r' && ptr < end))
+      {
+          ++ptr;
+      }
+    }
+  } while (!found);
+
```
Jamiras, line 619: there should be an `rc_file_close` here.

src/rhash/hash.c
```diff
@@ -868,6 +880,105 @@ static int rc_hash_pcfx_cd(char hash[33], const char* path)
...
+  last_track_lba = rc_cd_get_lba(last_track_handle);
+
+  /* its offset being 45000 means it is track 03.
+     if so, start finding executable from here. */
+  if (last_track_lba == 45000)
+    sector = rc_cd_find_file_sector(last_track_handle, exe_file, &size, last_track_lba);
+  else
+    sector = rc_cd_find_file_sector(track_handle, exe_file, &size, rc_cd_get_lba(track_handle));
+  
+  rc_cd_close_track(track_handle);
```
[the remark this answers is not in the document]
Jamiras: I believe the functionality is correct - the TOC will always be in track 3, and the boot code will be in the last track. Suggest separating the functionality to make this more clear:
```
sector = rc_cd_find_file_sector(track_handle, exe_file, &size, rc_cd_get_lba(track_handle));
rc_cd_close_track(track_handle);

if (sector == 0)
  return rc_hash_error("Could not locate boot executable");

/* last track contains the boot executable */
last_track_handle = rc_cd_open_track(path, rc_cd_num_tracks(path));
last_track_lba = rc_cd_get_lba(last_track_handle);
sector -= last_track_lba;
```

src/rhash/hash.c
```diff
@@ -868,6 +880,105 @@ static int rc_hash_pcfx_cd(char hash[33], const char* path)
...
+
+  /* subtract it by its LBA to get the true sector location on the file */
+  sector -= last_track_lba;
+
+  if (!sector)
+    rc_hash_error("Could not locate boot executable");
```
Jamiras, line 953: this check needs to occur before subtracting `last_track_lba` and should exit the function, as `num_read` and `buffer` will be invalid going into the loop.

include/rhash.h
Jamiras: Your "really high number" is a negative number (-105,840,000), and is effectively removing the 45000 sectors from the directory entries. As I mentioned, the one caveat to this is the TOC sector (that's normally in sector 16) has to be offset in the other direction. The 16th sector of track 3 is sector 45016, so that's where the TOC should be read from. If you look for the TOC in sector 45016, the calculation becomes:
```
sector_start = 45016 * 2352 + 16 + (-105840000)
sector_start = 105877632 + 16 - 105840000
sector_start = 37648
```
resulting in a positive offset in the track 3 file adjusted for the lba.

Similarly, when the TOC says the boot file is in sector 45020, sector_start comes out to 47056.

The meta information can be similarly located at sector 45000.

miagui: I understand now, that's a pretty good solution!
I've now just commited all changes mentioned in the code review.

Jamiras (approved):

PR #115, opened 2021-02-16
Remove defunct Lua section from readme

Sanaki:
Since Lua support is no longer being considered, removing this section would help avoid confusion from developers interested in incorporating the library into their clients.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,20 +8,6 @@ Not all structures defined by **rcheevos** can be created via the public API, bu
 
 Finally, **rcheevos** does *not* allocate or manage memory by itself. All structures that can be returned by it have a function to determine the number of bytes needed to hold the structure, and another one that actually builds the structure using a caller-provided buffer to bake it.
 
-## Lua
-
-RetroAchievements is considering the use of the [Lua](https://www.lua.org) language to expand the syntax supported for creating achievements. The current expression-based implementation is often limiting on newer systems.
-
-At this point, to enable Lua support, you must compile with an additional compilation flag: `HAVE_LUA`, as neither the backend nor the UI for editing achievements are currently Lua-enabled.
-
-> **rcheevos** does *not* create or maintain a Lua state, you have to create your own state and provide it to **rcheevos** to be used when Lua-coded achievements are found. Calls to **rcheevos** may allocate and/or free additional memory as part of the Lua runtime.
-
-Lua functions used in trigger operands receive two parameters: `peek`, which is used to read from the emulated system's memory, and `userdata`, which must be passed to `peek`. `peek`'s signature is the same as its C counterpart:
-
-```lua
-function peek(address, num_bytes, userdata)
-```
-
 ## API
 
 > An understanding about how achievements are developed may be useful, you can read more about it [here](http://docs.retroachievements.org/Developer-docs/).


Sanaki:
Looking this over, seems it'll need a bit more cleanup than just removing the main section.

meleu:
Do you have any information from the libretro guys (or even @Jamiras) saying that they won't support Lua? (if yes can you provide a link?)

Although it's not a priority, I still have interest in supporting Lua for achievement creation some day.

Ben-Ghazi:
KickMeElmo said that lua support has been rejected on discord.

https://discord.com/channels/310192285306454017/310195854642249728/806375287570956328

Sanaki:
I am KickMeElmo. That said, I went looking for a public source. Seems it wasn't explicitly rejected as I had remembered, but Jamiras has stated he's against supporting it. More may have been said privately, but for now, this is a source for the aforementioned statement:
https://discord.com/channels/310192285306454017/645777658319208448/724835036049833985

Without severely limiting what can be done, there are serious security implications that come from incorporating Lua support. As much as I'd adore it, we'd run the risk of allowing a rogue dev to deliver a malicious payload to ever player of any given set.

leiradel:
> Without severely limiting what can be done

Lua can be sandboxed to avoid security issues, like removing access to the `io` and `os` modules. It would still be possible to run an infinite loop or to create a string that consumes all the available memory, but these won't cause any permanent problems.

meleu:
Those concerns about Lua were brought before but never discussed in a centralized place. I suggest we use this PR thread for it.

---

Adding here the [Jamiras' words](https://discord.com/channels/310192285306454017/645777658319208448/724835036049833985) to help with context:

> I'm actually mostly against it. We can't control whatever code the devs put into lua. They could easily write an infinite loop or otherwise affect the performance of the game.
>
> The main thing it would provide is state, and there's other ways to do that. Variables and multi-session would solve almost everything lua would
>
> It was also going to be a solution for pointers, but we managed to find another way to handle that.
Besides, most devs don't know programming languages. I've seen some of the stuff that comes through jr. devs, and I'd be worried to see what they'd do with lua.

---

I agree with the security concerns. The most blatant ones would be solved by what @leiradel mentioned above.

And regarding the infinite-loop problem, this concern was brought in the beginning of rcheevos implementation. At that time I did some research about how to disable loops and found a promising thing in the Lua mailing list archive.

In [this message](http://lua-users.org/lists/lua-l/2014-03/msg00447.html) there's a guy submitting a patch implementing a method to "unreserve" some reserved words. And then one of the designers of the Lua language [proposed a more elegant way to achieve that](http://lua-users.org/lists/lua-l/2014-03/msg00449.html).

I didn't go to far into that because we, at RetroAchievements, have many other problems to solve before Lua adoption, specially in the backend server. Problems that could even be aggravated by premature Lua adoption.

Credit: I found those links via [this stackoverflow question](https://stackoverflow.com/questions/28365431/lua-can-i-disable-parts-of-the-language-from-c)

---

Regarding current achievement creators not knowing programming, this is not a valid point against Lua adoption. The current method will be kept available.

In the "people" side of things, the main advantage is that achievement creation will be a more programmer-friendly activity and I'm sure we'll get many nerds onboard, creating amazing things.

leiradel:
> I agree with the security concerns. The most blatant ones would be solved by what @leiradel mentioned above.

[Here](https://github.com/kikito/sandbox.lua/blob/master/sandbox.lua#L44)'s a more comprehnsive list.

> And regarding the infinite-loop problem, this concern was brought in the beginning of rcheevos implementation. At that time I did some research about how to disable loops and found a promising thing in the Lua mailing list archive.

I'm against a heavy hand solution like this, and the link above is a complete sandbox solution that will interrupt a script if it loops too much. However, to count loop iterations a debug hook must be used and this slows down Lua execution, so every legitimate loop would be impacted.

I think that with great power comes great responsibility :) Trolls that do nasty things on purpose will be spotted and banned, and inexperienced programmers will improve with help from others, specially feedback about their code.

> In the "people" side of things, the main advantage is that achievement creation will be a more programmer-friendly activity and I'm sure we'll get many nerds onboard, creating amazing things.

This is the most compelling reason to add Lua scripting. Retro Achievements could expand into a modding community more or less like [what I wrote here](https://gist.github.com/leiradel/a04e85df397d65a6ae12617c23339b13).

meleu:
> (...) so every legitimate loop would be impacted.

We don't have anything like a loop in the current method of achievement creation (and the creators were still able to produce some really awesome work).

In fact, loops being started in every single frame of a game isn't something we need, at all.

> Trolls that do nasty things on purpose will be spotted and banned, and inexperienced programmers will improve with help from others, specially feedback about their code.

This is exactly what I want to avoid. I mean, I want to remove the human-factor when it's possible.

That would put a responsibility in people who are just volunteers.

We currently see achievement sets from newbies being accumulated waiting for a reviewer. I think adding "Lua programming" AND "Lua safe programming" to the skillset needed to be a reviewer, would reduce the number of reviewers and make that queue even longer.

Also, the social interactions between the parts involved are not always as smooth as we'd like. And sometimes we see skilled people leaving the project just because of poor interactions.

So, I think the more boring/repetitive tasks we can delegate to machines, the better.

leiradel:
> In fact, loops being started in every single frame of a game isn't something we need, at all.

I think adding a cut-down language doesn't add much value, it won't allow people to do much more than what's already possible with the current method.

meleu:
You're right... After a second thought I realized I was hasty in my statement.

We can take advantage of loops, for example, to go through an array of address to scan certain parts of the emulated memory...

That repo you linked looks very helpful for what we want: https://github.com/kikito/sandbox.lua

In their README they say they are:

> A pure-lua solution for running untrusted Lua code.
> 
> The default behavior is restricting access to "dangerous" functions in Lua, such as `os.execute`.
> 
> It's possible to provide extra functions via the `options.env` parameter.
> 
> Infinite loops are prevented via the `debug` library.

Please, correct me if I'm wrong: that kind of thing would need to be done in the RetroArch code, right?

Jamiras:
This is only removing the references from the documentation. It doesn't do anything to the code. I'd prefer both remain in sync - whether we decide to gut LUA entirely, or just pretend it's not there.

As for actually supporting LUA - my thoughts have been captured above. Some of our achievement sets are already getting complicated enough that they're starting to impact performance on lower end systems that are already taxed by trying to emulate the more advanced systems where the complex logic is required. LUA may simplify that logic, but unless the coder is very proficient, it will likely be less performant.

For one example, the runtime is able to identify memory references shared between conditions (and even between achievements) and only query the emulator for each value once. Each of these fetched values is pointed to by the conditions that use it, so any overhead of redundant lookups is completely mitigated.

There are additional concerns too:
* What does the LUA editor look like? We'd probably have to treat it like Rich Presence where we just monitor an external file to pick up changes and have that uploaded to the server indirectly through the website.
* How do we capture LUA state in a save state? LUA probably provides a way to do this, but I'm not sure.
* What happens if LUA generates an error? Do we treat it as if the function returned 0, do we ignore the achievement for this frame, or do we mark it as unsupported until the user reloads the game? Do we report the error to the user?

leiradel:
> How do we capture LUA state in a save state? LUA probably provides a way to do this, but I'm not sure.

Lua doesn't provide any facilities to capture the state of a running program. There are heavy-duty serialization libraries like Pluto, but it's too intrusive IMO. The best way would be to look for serialization and deserialization functions in the Lua script, but that will make things even harder for the casual achievement developer.

> What happens if LUA generates an error?

I don't see this particular problem as a show stopper, if the Lua script generates an error, the achievement has an error and has to be fixed. It should be disabled until there's a new version, and the error should be reported back.

While I'd love to provide a programming language for achievement developers, I think Lua is not the right choice. Maybe the right answer here would be a [custom language](https://github.com/leiradel/Retrophies), where we're in control of everything including state saving and loading. The drawback is of course the maintenance cost of the language spec, compiler, VM, and integration.

meleu:
> Maybe the right answer here would be a custom language

The language used by [Jamiras' RATools](https://github.com/Jamiras/RATools/wiki/Script-File-Syntax) seems to be a good candidate. And it's already being used by some achievement creators.

leiradel:
> The language used by [Jamiras' RATools](https://github.com/Jamiras/RATools/wiki/Script-File-Syntax) seems to be a good candidate

AFAIK it's not a programming language, just a different way to write the memrefs, but if it's enough for the developers I agree.

Sanaki:
Given that this isn't going to be merged, I'm closing this PR.

PR #119, opened 2021-03-18
Ignore 7800 header if detected

Sanaki:
I essentially cloned the PCE header skip method for this. The smallest game I found was 4KiB and I verified all other games were a multiple of that value. Atari 7800 headers do have a detectable string that can be used if preferred, but I wasn't sure how to go about doing that properly. If desired, this is the No-Intro header skipper method:
```xml
<rule start_offset="80" end_offset="EOF" operation="none">
	<data offset="1" value="415441524937383030" result="true"/>
	<data offset="60" value="0000000041435455414C20434152542044415441205354415254532048455245" result="true"/>
```
Details of the header can be found here:
http://7800.8bitdev.org/index.php/A78_Header_Specification

### DIFF
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -502,9 +502,8 @@ static int rc_hash_3do(char hash[33], const char* path)
 
 static int rc_hash_7800(char hash[33], uint8_t* buffer, size_t buffer_size)
 {
-  /* if the file contains a header, ignore it (expect ROM data to be multiple of 4KB) */
-  uint32_t calc_size = ((uint32_t)buffer_size / 0x1000) * 0x1000;
-  if (buffer_size - calc_size == 128)
+  /* if the file contains a header, ignore it */
+  if (memcmp(&buffer[1], "ATARI7800", 9) == 0)
   {
     rc_hash_verbose("Ignoring 7800 header");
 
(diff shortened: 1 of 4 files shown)


Jamiras:
It's always preferable to match a magic string than to rely on file size. You're much less likely to have a conflict when checking the magic string.

The first rule:
```
<data offset="1" value="415441524937383030" result="true"/>
```
Is matching the "ATARI7800" string at 1 byte into the file. That seems like it would be sufficient enough for the check. Just replace the
```
  uint32_t calc_size = ((uint32_t)buffer_size / 0x1000) * 0x1000;
  if (buffer_size - calc_size == 128)
```
with
```
   if (memcmp(&buffer[1], "ATARI7800", 9) == 0)
```
and you should be good-to-go.

The second rule is trying to match the "ACTUAL CART DATA STARTS HERE", which makes it even less likely to get a false positive, but I don't feel it's necessary.

I would also like to see some tests added to `test_hash.c`, but I can add those later.

Sanaki:
Updated. For now, I left the double test present, though as you said, a better test would be preferable considering.

Jamiras (approved):
~~I have improved tests ready-to-go once this is merged. Cannot push directly to the PR as the source is your repository.~~
Tests have been added. Just had to add your fork as a remote.

PR #120, opened 2021-03-24
provide fully-encapsulated API for sever interaction

Jamiras:
Provides a replacement for the existing `rurl` functionality, which will be deprecated by this solution. A series of API calls with structured inputs and outputs is provided, eliminating the need for the called to do their own JSON parsing. The caller is still responsible for their own HTTP calls.

Each API is structured this way - `rc_api_XXX_request_t` -> `rc_api_request_t` -> `rc_api_XXX_response_t` and performed in five steps.

1) The API parameters are initialized by zero-ing out a structure and then populating the relevant fields. This allows expanding the inputs in the future without requiring new API definitions.
```
    rc_api_award_achievement_request_t api_params;
    memset(&api_params, 0, sizeof(api_params));

    api_params.username = "Username";
    api_params.api_token = "API_TOKEN";
    api_params.achievement_id = 12345;
    api_params.hardcore = 0;
```

2) The API parameters are turned into an HTTP request.
```
    rc_api_request_t api_request;
    rc_api_init_award_achievement_request(&api_request, &api_params);
```

3) The request is sent to the server. The client is responsible for providing code to do this. The `rc_api_request_t` object has a URL and POST data to post. A minor concern with the `rurl` set of functions is that they pass everything (including sensitive information) as GET parameters.

4) The response is converted into data structures.
```
    rc_api_award_achievement_response_t api_response;
    const auto result = rc_api_process_award_achievement_response(&api_response, http_response_content);
    if (result != RC_OK)
    {
        /* parse error, check for HTTP error codes */
        if (http_response_code != 200)
        {
            /* handle HTTP error */
        }
        else
        {
            /* handle rcheevos error */
        }
    }
    else if (!api_response.response.succeeded)
    {
        /* server provided error (in api_response.response.error) */
    }
    else
    {
        /* success. return updated player score*/
        new_player_score = api_response.new_player_score;
    }
```

5) The request and response structures are destroyed. Memory may be allocated when converting input parameters to an HTTP request, and when converting JSON to response data.
```
    rc_api_destroy_request(&api_request);
    rc_api_destroy_award_achievement_response(&api_response);
```

This PR exposes all of the APIs required for retroarch:
* `login`
* `resolve_hash`
* `fetch_game_data`
* `start_session`
* `fetch_user_unlocks`
* `ping`
* `award_achievement`
* `submit_lboard_entry`

An additional API is provided for downloading badges. It does not require step 4 as the response is the raw image file.
* `fetch_image`

Support is also present for custom hosts (primarily to allow specifying the use of a non-https version of the site for RetroArch versions that don't support SSL). The default host in this PR is https.


leiradel:
It's usual to have a field, the first field, hold a value that informs the callee the version of the structure the caller is passing. Some use an incremental version number, others set it to the size of the structure.

Jamiras:
I don't like the "size of structure" versioning semantic as it causes the client to immediately claim to support the newest version every time it's compiled, regardless of whether or not changes were made to support the newest version. For example, if a new required field is added to a structure, the size increases, and the client says "I'm using the newest size", but doesn't pass the required field. The API then has to check for the presence of the field anyway, regardless of the version indicator. As such, the version indicator can just be ignored and the presence of the required field provides the versioning information. Additionally, the size can change depending on the compile flags (x86 vs x64, debug vs release) and field alignment/packing can allow some changes without actually affecting the size.

The `memset()` ensures all new fields are explicitly set to 0 (or NULL), so as long as the default of the new field can be 0 (or NULL), this API is forward compatible without any explicit versioning. If there ever is a case where a new field can be 0 (or NULL) and the expected default is non-zero, then an additional field could be added at that point with a non-zero value indicating the field was explicitly set to 0, including (but not limited to) a version field set to "1" or something similar.

leiradel:
Cool, "zero is the default" really sounds better. Is it possible that the client have a prior version of the structure and thus `memset`s only part of it?

Jamiras:
Right now, all clients that I'm aware of are compiling rcheevos directly into their code, so there should be no way for the library to get ahead of the client.

Versioning does become important when the library can get ahead of the client, so it doesn't try to read the fields that don't yet exist.

PR #121, opened 2021-03-24
add new console enums

televandalist:
Watara Supervision, Sharp X1, TIC-80, and Thomson TO8


leiradel:
Note: I've recently helped the author of the TIC-80 port to Libretro to fix the memory maps so developers of TIC-80 games can work with achievement developers to make achievements for TIC-80 games.

If anyone from either side needs some guidance I'm available to give a hand.

Jamiras (approved):

PR #130, opened 2021-05-08
ignore extra whitespace between fields in gdi file

Jamiras:


Sanaki (approved):
With these changes, the entire Redump Dreamcast GD-ROM set hashes successfully. MIL-CD is not supported yet, but would require an entirely different hashing approach.

PR #135, opened 2021-05-29
Add external RAM to memory map of Magnavox Odyssey 2

gdeOo:
The Magnavox Odyssey 2 has internal and external RAMs, which are accessed using unique instructions, instead of being mapped to a single address space.

The o2em core used to only expose the internal RAM, but many (maybe most?) games make use of the external RAM. Having no access to it is problematic for achievement development, so the core has been updated to expose the external RAM as `RC_MEMORY_TYPE_SYSTEM_RAM` right after the internal RAM, creating a virtual memory map that only exists for this purpose.

This commit updates the memory map in rcheevos accordingly.

Corresponding core PR has been merged: https://github.com/libretro/libretro-o2em/pull/47


Jamiras:
There's a unit test that asserts the amount of memory exposed by each system. I expect the CI to report it once I kick it off.

https://github.com/RetroAchievements/rcheevos/blob/74631ea6779cec15c0371ef4ff17331c025d1a0f/test/rcheevos/test_consoleinfo.c#L127

gdeOo:
Forgot to run the tests locally 😞 Updated the test.

Jamiras (approved):

PR #143, opened 2021-07-04
Allow using https:// with RCHEEVOS_URL_PROTOCOL

RobLoach:
This change introduces a `RCHEEVOS_URL_PROTOCOL` define which allows changing the requests to use `https://` instead of `http://`. Inspired by https://github.com/libretro/RetroArch/pull/11551 .

### DIFF
--- a/src/rurl/url.c
+++ b/src/rurl/url.c
@@ -6,6 +6,10 @@
 #include <stdio.h>
 #include <string.h>
 
+#ifndef RCHEEVOS_URL_PROTOCOL
+#define RCHEEVOS_URL_PROTOCOL "http://"
+#endif
+
 static int rc_url_encode(char* encoded, size_t len, const char* str) {
   for (;;) {
     switch (*str) {
@@ -67,7 +71,7 @@ int rc_url_award_cheevo(char* buffer, size_t size, const char* user_name, const 
   written = snprintf(
     buffer,
     size,
-    "http://retroachievements.org/dorequest.php?r=awardachievement&u=%s&t=%s&a=%u&h=%d",
+    RCHEEVOS_URL_PROTOCOL"retroachievements.org/dorequest.php?r=awardachievement&u=%s&t=%s&a=%u&h=%d",
     urle_user_name,
     urle_login_token,
     cheevo_id,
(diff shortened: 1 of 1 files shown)


CookiePLMonster:
DuckStation also does the same, e.g. https://github.com/stenzek/duckstation/blob/master/dep/rcheevos/src/rurl/url.c#L70

RobLoach:
Thinking about this, would it be better to have this managed as a static variable or a passed in boolean? That way it wouldn't require recompilation in order to use HTTPS.

CookiePLMonster:
I don't think this needs to be decided on runtime, as in your application you either can submit HTTPS requests or you can't.

Jamiras:
It would probably be easier to just have a `RCHEEVOS_SSL_HOST=1` #define than expecting the client to specify the entire protocol - especially from the command line. I'm not sure if the "://" would need to be escaped, or how easy that would be to do. Alternately, the define could just be "http" or "https" and the "://" would be left in each of the strings.

`rurl` is deprecated in rcheevos 10.0, but I'm fine with including this change anyway (with one of the suggested tweaks above).

The `rapi` suite uses the SSL host by default and provides a runtime function ([rc_api_set_host](https://github.com/RetroAchievements/rcheevos/wiki/rc_api_set_host)) to specify a custom host, which can be used to target the non-SSL host. rcheevos 10.0 was published a month and a half ago, and I do plan to update RetroArch soon. I just haven't had a chance yet.

Note that the host requires TLS 1.2+, which is not enabled by default on Windows 7 or older versions (see https://github.com/RetroAchievements/RAIntegration/pull/721), so there may still be a need to have a runtime toggle.

src/rurl/url.c
```diff
@@ -158,7 +162,7 @@ int rc_url_get_badge_image(char* buffer, size_t size, const char* badge_name) {
   int written = snprintf(
     buffer,
     size,
-    "http://i.retroachievements.org/Badge/%s",
+    RCHEEVOS_URL_PROTOCOL"i.retroachievements.org/Badge/%s",
```
Jamiras, line 165: The image host doesn't currently support HTTPS. This one shouldn't change.

RobLoach:
``` suggestion
    "http://i.retroachievements.org/Badge/%s",
```

RobLoach: Thanks!

RobLoach:
Nice change! 👍

RobLoach:
`RCHEEVOS_SSL_HOST` makes sense to me.

src/rurl/url.c
```diff
@@ -6,6 +6,10 @@
 #include <stdio.h>
 #include <string.h>
 
+#ifndef RCHEEVOS_URL_PROTOCOL
+#define RCHEEVOS_URL_PROTOCOL "http"
+#endif
```
RobLoach, line 11: If we want to go the `RCHEEVOS_SSL_HOST` route, would it look like this?
``` suggestion
#ifndef RCHEEVOS_URL_PROTOCOL
#ifdef RCHEEVOS_SSL_HOST
#define RCHEEVOS_URL_PROTOCOL "https"
#else
#define RCHEEVOS_URL_PROTOCOL "http"
#endif
#endif
```

Jamiras: I'd have left the "#ifndef RCHEEVOS_URL_PROTOCOL" out. it's not like there's any other reasonable choices.
```
#ifdef RCHEEVOS_SSL_HOST
#define RCHEEVOS_URL_PROTOCOL "https"
#else
#define RCHEEVOS_URL_PROTOCOL "http"
#endif
```

RobLoach: Sounds good.

PR #145, opened 2021-07-13
add rapi calls for fetch_leaderboard_info, fetch_achievement_info, and fetch_game_list

Jamiras:
supports new `lbinfo` changes (https://github.com/RetroAchievements/RAWeb/pull/627)


CookiePLMonster:
Any chance of seeing `lbinfo` changes added to `rurl` too? `rapi` is overkill if response parsing is already done on the app side.

Jamiras:
> rapi is overkill if response parsing is already done on the app side.

The point of deprecating **rurl** was to eliminate the need for the app to do the parsing as the actual data structures passed through the API are not documented and the client shouldn't need to decipher them. **rurl** served as a good first step by abstracting the requests, but did nothing for abstracting the response. **rapi** does that.

As duckstation is already effectively implementing their own **rurl** functions for `lbinfo`, I've added two official functions to **rurl**: `rc_url_get_lboard_entries` and `rc_url_get_lboard_entries_near_user`.

CookiePLMonster:
> As duckstation is already effectively implementing their own **rurl** functions for `lbinfo`, I've added two official functions to **rurl**: `rc_url_get_lboard_entries` and `rc_url_get_lboard_entries_near_user`.

Looks good, with just one nitpick - isn't an offset parameter valid for near_user requests too? If I recall correctly from RAWeb, it is valid and it should allow offsetting the request relatively to the user - and thus it could be negative too.

PR #150, opened 2021-08-07
Add 16-bit/24-bit/32-bit big-endian memory sizes

Jamiras:
Supports https://github.com/RetroAchievements/RAIntegration/issues/200

While we don't yet support any natively big-endian systems, there are a few use cases where being able to read multiple bytes in big-endian order would be convenient.

The one listed in the referenced issue is for displayed scores. When "01 05 00 00" appears in memory, but means "1500", reading it as little endian would generate something akin to "0051", so to check for "score > 1500", you'd have to check for the first byte being >= 2, or the first byte being 1 and the second byte being >= 5 or the first two bytes being 1 and 5 and the third or fourth bytes being greater than 0. Reading the value as a 32-bit big-endian value rearranges everything so you can just do a single check for > 1500.

Another potential application is Applesoft Real Numbers (5-byte floating point numbers on the Apple II heavily used in Oregon Trail). The first byte is the exponent, and the next four bytes are a big-endian mantissa. For positive numbers, using a big endian read of the first four of the five bytes will generate a value that can be used for relative comparisons. For example to check if a value is greater than 10.0, you could do: "32-bit BE $ADDR >= 0x84200000", where the actual encoding for 10.0 is `84 20 00 00 00`. Actually decoding the float could potentially be supported at a later time if we ever support floating point logic.

By making it a memory size, you can still use normal flags, like delta, and bcd-decoding.

Only 16-bit, 24-bit, and 32-bit sizes are supported in big-endian mode. Anything 8-bit or less is unaffected by endianness.


Blazekickn:
24bit big endian is more common than you think, i've seen it in a couple games. Gen 2 Pokemon uses it to store money for example (shown below)

image

I had to code an achievement in one of the romhacks for having over a certain amount of money once and while it was possible, it was really annoying to do and required alts. Having big endian for that would've been nice.

PR #154, opened 2021-08-13
address compiler warnings

Jamiras:
Fixes more warnings captured in some of the RetroArch builds, as well as several unused variable and casting-type warnings generated when using /w4 in visual studio.
```
In function 'rc_url_builder_append',
    inlined from 'rc_url_builder_append' at deps/rcheevos/src/rapi/rc_api_common.c:862:6,
    inlined from 'rc_api_init_fetch_image_request' at deps/rcheevos/src/rapi/rc_api_common.c:1024:7:
deps/rcheevos/src/rapi/rc_api_common.c:864:5: warning: 'memcpy' forming offset [5, 8] is out of the bounds [0, 5] [-Warray-bounds]
  864 |     memcpy(builder->write, data, len);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
```
In file included from /Users/gitlab/builds/FkGsyN3z/2/libretro/RetroArch/griffin/griffin.c:198:
/Users/gitlab/builds/FkGsyN3z/2/libretro/RetroArch/griffin/../deps/rcheevos/src/rapi/rc_api_common.c:361:10: warning: implicit conversion loses integer precision: 'long' to 'unsigned int' [-Wshorten-64-to-32]
  return strtol(hex, NULL, 16);
  ~~~~~~ ^~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/rapi/rc_api_common.c
+++ b/src/rapi/rc_api_common.c
@@ -296,6 +296,7 @@ static int rc_json_missing_field(rc_api_response_t* response, const rc_json_fiel
 
 int rc_json_get_required_object(rc_json_field_t* fields, size_t field_count, rc_api_response_t* response, rc_json_field_t* field, const char* field_name) {
   const char* json = field->value_start;
+  (void)field_name;
 
   if (!json)
     return rc_json_missing_field(response, field);
@@ -357,6 +358,7 @@ int rc_json_get_required_array(unsigned* num_entries, rc_json_field_t* iterator,
     return rc_json_missing_field(response, field);
   }
 
+  (void)field_name;
   memcpy(iterator, field, sizeof(*iterator));
   ++iterator->value_start; /* skip [ */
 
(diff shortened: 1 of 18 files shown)


src/rapi/rc_api_common.c
```diff
@@ -388,7 +397,7 @@ static unsigned rc_json_decode_hex4(const char* input) {
   memcpy(hex, input, 4);
   hex[4] = '\0';
 
-  return strtol(hex, NULL, 16);
+  return (unsigned)strtol(hex, NULL, 16);
```
CookiePLMonster, line 400: Any reason not to use `strtoul`?

CookiePLMonster (commented):
For `int` -> `char` changes, it's better to specify `unsigned char` or `signed char` explicitly. Assumptions about the signedness of `char` shouldn't be made as some architectures (ARM) default to `char` being unsigned, while others treat it as signed.

Jamiras:
> For int -> char changes, it's better to specify unsigned char or signed char explicitly. 

Those changes just downcast to the same type as used in the variable the value is being stored in. Most of them are booleans (0 vs non-zero), and the others are enums (currently don't have enough values to reach sign conflicts).

> some architectures (ARM) default to char being unsigned

Per #149, one of the CI builds now uses `-fno-signed-char`.

src/rapi/rc_api_common.c
Jamiras: Nope. Hex numbers (base 16) can't be signed. And this function is specifically for decoding 16-bit hex numbers. The downcast to `unsigned [int]` is still required as `strtoul` returns `unsigned long`.

PR #155, opened 2021-08-18
Add hash algorithm for PSP discs

Jamiras:
Contents of `PSP_GAME\\PARAMS.SFO` and `PSP_GAME\\SYSDIR\\EBOOT.BIN` are concatenated together and hashed. `PARAMS.SFO` contains metadata about the disc (game name, version, serial number). `EBOOT.BIN` is the primary executable.


Sanaki (approved):
Approved based on hash testing across Redump and No-Intro PSP iso sets. Currently no support for bare pbp, which will eventually be desired for homebrew and bare PSN dumps.

PR #175, opened 2021-12-14
support for hashing dreamcast bin/cues

Jamiras:
Dreamcast cue files don't accurately represent the area between the low-density tracks and the high-density tracks.
```
REM SINGLE-DENSITY AREA
FILE "track01.bin" BINARY
  TRACK 01 MODE1/2352
    INDEX 01 00:00:00
    REM file is 705600 bytes (300 sectors)
FILE "track02.bin" BINARY
  TRACK 02 AUDIO
    INDEX 00 00:00:00
    REM two second pregap is 150 sectors, putting track 2 at sector 450 (correct)
    INDEX 01 00:02:00
    REM file is 1589952 bytes (676 sectors), putting track 3 at sector 1126 (should be 45000)
REM HIGH-DENSITY AREA
FILE "track03.bin" BINARY
  TRACK 03 MODE1/2352
    INDEX 01 00:00:00
```

Track 3 (the start of the high-density area) is expected to always start at sector 45000, but this isn't reflected in any of the cue files that I've seen for Dreamcast. Since the Dreamcast filesystem uses absolute sector addresses, this causes the code that locates the primary executable to not find the executable and not generate a hash.

The solution is to force track 3 to start at sector 45000 when generating a Dreamcast hash.

I've also made some additional changes to the cue processing code to better handle absolute sector to file location conversions for tracks with pregaps.

### DIFF
--- a/include/rc_hash.h
+++ b/include/rc_hash.h
@@ -101,7 +101,7 @@ extern "C" {
    */
   typedef void* (*rc_hash_cdreader_open_track_handler)(const char* path, uint32_t track);
 
-  /* attempts to read the specified number of bytes from the file starting at the read pointer.
+  /* attempts to read the specified number of bytes from the file starting at the specified absolute sector.
    * returns the number of bytes actually read.
    */
   typedef size_t (*rc_hash_cdreader_read_sector_handler)(void* track_handle, uint32_t sector, void* buffer, size_t requested_bytes);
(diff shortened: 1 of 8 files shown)


Sanaki (approved):
I found a failure to hash on one potentially valid game (`Alice Dreams (France) (Demo) (Unl)`), but given the breadth of other successfully hashed files, I don't believe this to be a flaw in the PR. Otherwise I haven't found any negative impact to hashing of other systems and hashing of both cue and gdi line up perfectly, with MIL-CD dumps hashing correctly as well. Beautiful work.

PR #204, opened 2022-09-22
Split PC Engine and PC Engine CD

televandalist:


src/rcheevos/consoleinfo.c
```diff
@@ -604,12 +607,20 @@ static const rc_memory_regions_t rc_memory_regions_pc8800 = { _rc_memory_regions
 /* ===== PC Engine ===== */
 /* http://www.archaicpixels.com/Memory_Map */
 static const rc_memory_region_t _rc_memory_regions_pc_engine[] = {
+    { 0x000000U, 0x001FFFU, 0x1F0000U, RC_MEMORY_TYPE_SYSTEM_RAM, "System RAM" },
+    { 0x012000U, 0x041FFFU, 0x0D0000U, RC_MEMORY_TYPE_SYSTEM_RAM, "Super System Card RAM" }
+};
+static const rc_memory_regions_t rc_memory_regions_pc_engine = { _rc_memory_regions_pc_engine, 4 };
```
Jamiras, line 613: There's only two entries not, so the `4` needs to be changed to `2`.

Additionally, removing the CD RAM block in the middle causes the RA addresses of the "Super System Card RAM" to change. The 0x012000U would have to be changed to `0x002000U`, and any games dependent on those addresses would have to have their code updated (which would break older versions of RetroArch that haven't received this change).

I thought the Super System Card was a CD-ROM expansion. In that case, it could be removed too.
> This was followed by the Super System Card released on October 26, an upgrade for the existing CD-ROM² add-on that serves as a replacement to the original System Card. 

If not, we'll probably need to put in a padding region to ensure the RA addresses are unchanged.

test/rcheevos/test_consoleinfo.c
```diff
@@ -154,6 +155,7 @@ void test_consoleinfo(void) {
   TEST_PARAMS2(test_memory, RC_CONSOLE_ORIC, 0x010000);
   TEST_PARAMS2(test_memory, RC_CONSOLE_PC8800, 0x011000);
   TEST_PARAMS2(test_memory, RC_CONSOLE_PC_ENGINE, 0x42800);
```
Jamiras, line 157: This value needs to be updated to reflect the lower amount of RAM exposed for non-CD PCE

src/rhash/hash.c
```diff
@@ -1918,6 +1918,7 @@ int rc_hash_generate_from_file(char hash[33], int console_id, const char* path)
     case RC_CONSOLE_MEGADUCK:
     case RC_CONSOLE_NEOGEO_POCKET:
     case RC_CONSOLE_ORIC:
+    case RC_CONSOLE_PC_ENGINE:
```
Jamiras, line 1921: This should be moved down into the same block as RC_CONSOLE_NINTENDO, as we also strip the header from PCE ROMs.

test/rcheevos/test_consoleinfo.c
televandalist: Changed to `0x02000`

Jamiras:
Looks like there's an invalid `RC_CONSOLE_PC_ENGINE` on line 939 of test_hash.c. If you change that to `RC_CONSOLE_PCFX`, it should clear up the remaining unit test.

Jamiras (approved):

PR #205, opened 2022-09-29
add RC_MEMSIZE_MBF32_LE

Jamiras:
Locomotive BASIC (on the CPC) uses MBF40 floats, but in little-endian format. We can't really read the full 40 bits, but the data structure does allow us to ignore the first byte and just use MBF32 on the other four. We still need to byte-swap the order though, so this PR adds a new MBF32_LE memory read. MBF32 will continue to be big endian.


redwizard42:
Almost every Amstrad CPC game I've looked at ran a .BAS file of the .dsk image, though so far I've only found one that [likely] is using floats. But since it is apparently the default data type for this BASIC, this should hopefully open up possibilities that weren't possible before.

Jamiras:
I would expect most games to be pre-compiled assembly that is loaded into memory and executed. BASIC interpretation is very slow. It's possible that the .BAS being loaded just bootstraps a compiled executable into memory and jumps into it.

leiradel:
One ZX Spectrum game I reversed engineered was mixed BASIC and machine code. The gameplay was implemented in machine code, but the intermission screen and bonus points were in BASIC. I had to read numbers from BASIC variables to implement the required functionality.

PR #207, opened 2022-10-18
expose additional 256KB of RAM for 32x

Jamiras:


televandalist:
I wonder if we'll be able to find something in Blackthorne :O

Jamiras:
Unfortunately, picodrive doesn't expose this extra memory, so you'll have to wait for a core/emulator that does.

PR #224, opened 2023-02-25
Added rc_hash_gamecube to hash Gamecube ISOs

LillyJadeKatrin:


src/rhash/hash.c
```diff
@@ -1143,6 +1143,76 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+  char* header_buffer = (char*)malloc(header_size);
+  rc_file_read(file_handle, header_buffer, header_size);
+  md5_append(&md5, header_buffer, header_size);
+  free(header_buffer);
+  // Hash main.dol executable
+  rc_file_seek(file_handle, dol_offset, SEEK_SET);
+  char* dol_buffer = (char*)malloc(dol_size);
+  rc_file_read(file_handle, dol_buffer, dol_size);
+  md5_append(&md5, dol_buffer, dol_size);
+  free(dol_buffer);
```
Jamiras, line 1209: Please add verbose callbacks indicating what was hashed:
```
  if (verbose_message_callback)
  {
    char buffer[128];
    snprintf(buffer, sizeof(buffer), "Hashing %u byte header", header_size);
    verbose_message_callback(buffer);
  }
```
and

```
  if (verbose_message_callback)
  {
    char buffer[128];
    snprintf(buffer, sizeof(buffer), "Hashing primary executable (%u bytes @ %u)", dol_size, dol_offset);
    verbose_message_callback(buffer);
  }
```

src/rhash/hash.c
```diff
@@ -1143,6 +1143,76 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+    be_buffer[3 - ix] = le_buffer[ix];
+  }
+  return *((uint32_t*)be_buffer);
+}
+
+static int rc_hash_gamecube(char hash[33], const char* path)
+{
+  md5_state_t md5;
+  void* file_handle;
+  const int BASE_HEADER_SIZE = 0x2440;
```
Jamiras, line 1162: You should have some way to identify that the file is indeed a GameCube disc. If you pass a PSX disc to this function, it'll still try to return a hash, and since the addresses you're using for sizes could have anything in them, it could cause huge memory allocations.

Is there some constant identifier in the app_loader_header or BootDOL stuff?

src/rhash/hash.c
```diff
@@ -1143,6 +1143,76 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+      dol_size = offset + size;
+  }
+
+  md5_init(&md5);
+  // Hash headers
+  rc_file_seek(file_handle, 0, SEEK_SET);
+  char* header_buffer = (char*)malloc(header_size);
+  rc_file_read(file_handle, header_buffer, header_size);
+  md5_append(&md5, header_buffer, header_size);
+  free(header_buffer);
```
Jamiras, line 1203: You should verify that malloc returned memory before proceeding. 
```
  if (!buffer)
  {
    rc_file_close(file_handle);
    return rc_hash_error("Could not allocate temporary buffer");
  }
  ```

File: src/rhash/hash.c [hunk 4]
```diff
@@ -1143,6 +1143,76 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+  return *((uint32_t*)be_buffer);
+}
+
+static int rc_hash_gamecube(char hash[33], const char* path)
+{
+  md5_state_t md5;
+  void* file_handle;
+  const int BASE_HEADER_SIZE = 0x2440;
+
+  file_handle = rc_file_open(path);
```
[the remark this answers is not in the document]
CasualPokePlayer: For the GC and Wii, you can make an assumption here regarding those factors. GC and Wii used special drives, they didn't support all the fancy formats you could find on CDs and such. They were limited to a single track, and did not support the extra metadata, so they were always 2048 sector ISOs. Dolphin goes under this assumption too.

src/rhash/hash.c
```diff
@@ -1143,6 +1143,76 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+static uint32_t rc_read_bigendian_u32_value(void* file_handle)
+{
+  char le_buffer[4];
+  char be_buffer[4];
+  rc_file_read(file_handle, le_buffer, 4);
+  for (int ix = 0; ix < 4; ix++)
+  {
+    be_buffer[3 - ix] = le_buffer[ix];
+  }
+  return *((uint32_t*)be_buffer);
```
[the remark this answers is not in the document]
CasualPokePlayer: I assume this is coming from some pattern in Dolphin, which is not endian safe (so in practice ever in the next 20 or so years this doesn't matter, only if Dolphin ever decides to become endian safe or some new emulator that supports a big endian machine pops up and supports rcheevos will this matter in practice). Not a reason to do this anyways, better to not rely on these assumptions.

src/rhash/hash.c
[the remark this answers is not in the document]
LillyJadeKatrin: ISO is the MVP at the moment. I definitely wish to add RVZ support down the road (especially since Dolphin devs so heavily emphasize that format) but there's still questions about whether the decompression happens inside rcheevos or externally in whichever emulator is calling it.

src/rhash/hash.c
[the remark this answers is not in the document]
LillyJadeKatrin: Gamecube and Wii are both big-endian while Dolphin, at least in the code I referenced to get the addresses and filesizes the hash algorithm needs, appears to be little-endian. That said, I agree with Jamiras, doing the math this way is probably preferable because it's unambiguous - though I'll still throw it in a function because I have to reuse it quite a few times.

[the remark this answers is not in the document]
leiradel: Compilers can optimize it for just a byteswap, but unfortunately that's not the case with MSVC: https://godbolt.org/z/5j76vTna5

On Windows you can use byteswap functions directly, if the function is called on tight loops it can help improve performace: https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/byteswap-uint64-byteswap-ulong-byteswap-ushort?view=msvc-170

If you need it to be cross-platform, this can help: https://stackoverflow.com/questions/41770887/cross-platform-definition-of-byteswap-uint64-and-byteswap-ulong

src/rhash/hash.c
```diff
@@ -1143,6 +1143,92 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+    snprintf(message, sizeof(message), "Hashing %u byte header", header_size);
+    verbose_message_callback(message);
+  }
+
+  // GetBootDOLOffset
+  uint32_t dol_offset = (buffer[0x420] << 24) | (buffer[0x421] << 16) | (buffer[0x422] << 8) | buffer[0x423];
+
+  // Iterate through the 7 main.dol code segments
+  rc_file_seek(file_handle, dol_offset, SEEK_SET);
+  rc_file_read(file_handle, buffer, MAX_BUFFER_SIZE);
```
Jamiras, line 1199: See previous comment about only allocating/reading what you need. I'd suggest reading the DOL table into a stack buffer, iterate it to find the largest size needed, allocate that block, and read each segment into that.

src/rhash/hash.c
```diff
@@ -1143,6 +1143,92 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+  for (size_t i = 0; i < 7; i++)
+  {
+    uint32_t offset = (buffer[0x0 + i * 4] << 24) | (buffer[0x1 + i * 4] << 16) | (buffer[0x2 + i * 4] << 8) | buffer[0x3 + i * 4];
+    uint32_t size = (buffer[0x90 + i * 4] << 24) | (buffer[0x91 + i * 4] << 16) | (buffer[0x92 + i * 4] << 8) | buffer[0x93 + i * 4];
```
Jamiras, line 1203: I noticed that several code segments had 0 size. You can skip the hash/log messages for those. You could also skip the seek/read if you follow my suggestion to not preload 64MB of data.

src/rhash/hash.c
```diff
@@ -1143,6 +1143,92 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+    (buffer[BASE_HEADER_SIZE + 0x15] << 16) |
+    (buffer[BASE_HEADER_SIZE + 0x16] << 8) |
+    buffer[BASE_HEADER_SIZE + 0x17];
+  uint32_t apploader_trailer_size =
+    (buffer[BASE_HEADER_SIZE + 0x18] << 24) |
+    (buffer[BASE_HEADER_SIZE + 0x19] << 16) |
+    (buffer[BASE_HEADER_SIZE + 0x1a] << 8) |
+    buffer[BASE_HEADER_SIZE + 0x1b];
+  uint32_t header_size = BASE_HEADER_SIZE + apploader_header_size + apploader_body_size + apploader_trailer_size;
+
```
Jamiras, line 1183: Consider aborting if `header_size` exceeds MAX_BUFFER_SIZE (or some other reasonable value - 1MB?)

src/rhash/hash.c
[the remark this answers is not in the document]
Jamiras: We'd need to take things a step farther and determine if a byteswap was needed in the first place. Using the explicit shifts ensure we translate the data from the ROM/ISO correctly, regardless of the endianness of the host system. 

And I'm not terribly worried about the performance for pulling a couple dozens integers out of the file.

src/rhash/hash.c
```diff
@@ -1143,6 +1143,113 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+  // GetBootDOLOffset
+  // Base header size is guaranteed larger than 0x423 therefore buffer contains dol_offset right now
+  uint32_t dol_offset = (buffer[0x420] << 24) | (buffer[0x421] << 16) | (buffer[0x422] << 8) | buffer[0x423];
+  free(buffer);
+
+  // Find offsets and sizes for the 7 main.dol code segments and 11 main.dol data segments
+  uint8_t addr_buffer[0xD8];
+  uint32_t dol_offsets[18];
+  uint32_t dol_sizes[18];
+  uint32_t dol_buf_size = 0;
```
Jamiras, line 1204: The c89 build is going to fail on these. All variables have to be declared at the top of the function.

src/rhash/hash.c
```diff
@@ -1143,6 +1143,113 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+  // Iterate through the 18 main.dol segments and hash each
+  buffer = (uint8_t*)malloc(dol_buf_size);
+  if (!buffer)
+  {
+    rc_file_close(file_handle);
+    return rc_hash_error("Could not allocate temporary buffer");
+  }
+  for (uint32_t ix = 0; ix < 18; ix++)
+  {
+    if (dol_sizes[ix] == 0) continue;
```
Jamiras, line 1231: Please put `continue;` on a separate line

test/rhash/test_hash.c
```diff
@@ -804,6 +804,21 @@ static void test_hash_dreamcast_cue()
 
 /* ========================================================================= */
 
+static void test_hash_gamecube()
+{
+  size_t image_size;
+  uint8_t* image = generate_gamecube_iso(32, &image_size);
+  char hash[33];
+  int result = rc_hash_generate_from_buffer(hash, RC_CONSOLE_GAMECUBE, image, image_size);
```
Jamiras, line 812: gamecube should never be buffered. mock the file and call `rc_hash_generate_from_file`

```
  const char* filename = "test.iso";
  mock_file(0, filename, image, image_size);
  int result = rc_hash_generate_from_file(hash, RC_CONSOLE_GAMECUBE, filename);
```

LillyJadeKatrin: Okay, but I'm still trying to wrap my head around quite how to implement compression. If it's not in rcheevos then my inclination is to pass the decompressed buffer in, but I know you've mentioned CD readers in the past so I'll look into that.

src/rhash/hash.c
```diff
@@ -1143,6 +1143,113 @@ static int rc_hash_nintendo_ds(char hash[33], const char* path)
...
+  rc_file_seek(file_handle, 0x1c, SEEK_SET);
+  rc_file_read(file_handle, quad_buffer, 4);
+  if (quad_buffer[0] != 0xC2|| quad_buffer[1] != 0x33 || quad_buffer[2] != 0x9F || quad_buffer[3] != 0x3D)
+  {
+    rc_file_close(file_handle);
+    return rc_hash_error("Not a Gamecube disc");
+  }
+
+  // GetApploaderSize
+  rc_file_seek(file_handle, BASE_HEADER_SIZE, SEEK_SET);
```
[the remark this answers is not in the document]
LillyJadeKatrin: Guh, work habits. My office build system runs tests automatically so I'm too used to reading "builds" = "tests pass". I'll retrain myself on this.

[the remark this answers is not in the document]
LillyJadeKatrin: That said, I do believe the hash does need updating even with this fixed, on account of the uninitialized memory you mentioned in a previous comment that I've now removed.

src/rhash/hash.c
```diff
@@ -1859,6 +1970,7 @@ int rc_hash_generate_from_buffer(char hash[33], int console_id, const uint8_t* b
     case RC_CONSOLE_NINTENDO_64:
     case RC_CONSOLE_NINTENDO_DS:
     case RC_CONSOLE_NINTENDO_DSI:
+    case RC_CONSOLE_GAMECUBE:
```
Jamiras, line 1973: Gamecube should not be supported for buffer hashing

LillyJadeKatrin: D'oh! Forgot to remove this, thank you!

PR #227, opened 2023-03-03
add hash and memory map for TI-83

Jamiras:


src/rcheevos/consoleinfo.c
```diff
@@ -785,6 +785,15 @@ static const rc_memory_region_t _rc_memory_regions_thomson_to8[] = {
 };
 static const rc_memory_regions_t rc_memory_regions_thomson_to8 = { _rc_memory_regions_thomson_to8, 1 };
 
+/* ===== TI-83 ===== */
+/* https://wikiti.brandonw.net/index.php?title=83Plus:Memory_Mapping#Naming_Conventions */
+/* TODO: bizhawk exposes 256KB of memory, but it never changes, so it's hard to tell if it's correct */
```
CasualPokePlayer, line 790: BizHawk currently only emulates the original TI-83, not any other the later models (e.g. TI-83+). There's only 32KiB of RAM there, and it's not banked.

Jamiras: That's what Google was telling me, so why does bizhawk load 256KB into the RAIntegration DLL? And possibly more importantly, why doesn't any of that 256KB change when a game is being played?

CasualPokePlayer: I suspect it's falling back onto default handling which is load up the "main" memory domain (which I suspect here is ROM, which is a bug anyways, it should be RAM), so you don't get any changes.

Jamiras: Bizhawk Hex Editor opens at "TI83 ROM - 0x40000 addresses", which corresponds to 256KB, so you're probably right. Switching to RAM does only report 0x8000 addresses, so I should only expect to see that in rcheevos instead of the entire memory space?

PR #228, opened 2023-03-03
add hash and memory map for Uzebox

Jamiras:


src/rcheevos/consoleinfo.c
```diff
@@ -801,6 +801,13 @@ static const rc_memory_region_t _rc_memory_regions_tic80[] = {
 };
 static const rc_memory_regions_t rc_memory_regions_tic80 = { _rc_memory_regions_tic80, 10 };
 
+/* ===== Uzebox ===== */
+/* https://uzebox.org/index.php */
+static const rc_memory_region_t _rc_memory_regions_uzebox[] = {
+    { 0x000000U, 0x000FFFU, 0x000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "System RAM" }
```
Jamiras, line 807: The `uzem` core also exposes the EEPROM memory (1KB), but that's split into 32 blocks which can be allocated to any game. Even if there was something we wanted from the persistent data, finding it would be difficult. Assume the program also keeps a copy in RAM and don't expect to rely on the EEPROM data.

PR #230, opened 2023-03-11
Initialized byteswapped

LillyJadeKatrin:
While practically impossible, thanks to the return on line 795, in theory there is a code path that results in the byteswapped variable on line 810 of hash.c to be uninitialized when it is checked. The Dolphin cmake ninja compile path in particular throws a pedantic warning over it. This commit initializes it upon creation to a dummy value that will never get used but should be sufficient to satisfy any compilers.

### DIFF
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -750,7 +750,7 @@ static int rc_hash_jaguar_cd(char hash[33], const char* path)
   char message[128];
   void* track_handle;
   md5_state_t md5;
-  int byteswapped;
+  int byteswapped = 0;
   unsigned size = 0;
   unsigned offset = 0;
   unsigned sector = 0;


Jamiras:
It has nothing to do with the return on line 795. The compiler is seeing that the if/else chain inside the for loop has a path where byteswapped doesn't get initialized. That path would leave size at 0, which would then hit the return on line 795, but it's not the return specifically that's causing the warning.

src/rhash/hash.c
```diff
@@ -750,7 +750,7 @@ static int rc_hash_jaguar_cd(char hash[33], const char* path)
   char message[128];
   void* track_handle;
   md5_state_t md5;
-  int byteswapped;
+  int byteswapped = -1;
```
Jamiras, line 753: Recommend setting to 0. This is effectively a boolean, so -1 is invalid (and would evaluate as true).

LillyJadeKatrin:
Yeah no that was me overexplaining and poorly. I think we're on the same page.

Jamiras (approved):

PR #258, opened 2023-08-19
improve subsecond timer management

Jamiras:
Uses `clock_gettime` or `QueryPerformanceCounter` when available for sub-second event timings. (the logic for handling `clock` wrapping [may not have been correct](https://discord.com/channels/476211979464343552/757767535293890682/1138837836636885073)).

This also allows the client to provide their own implementation in case they already have a high-precision timer function, which actually makes testing easier as time elapsing can be simulated.

### DIFF
--- a/src/rcheevos/rc_client.c
+++ b/src/rcheevos/rc_client.c
@@ -13,7 +13,7 @@
 
 #ifdef _WIN32
 #define WIN32_LEAN_AND_MEAN
-#include <Windows.h>
+#include <windows.h>
 #include <profileapi.h>
 #else
 #include <time.h>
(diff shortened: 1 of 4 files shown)


src/rcheevos/rc_client.c
```diff
@@ -219,7 +219,7 @@ static rc_clock_t rc_client_clock_get_now_millisecs(const rc_client_t* client)
     return 0;
 
   /* round nanoseconds to nearest millisecond and add to seconds */
-  return (rc_clock_t)(now.tv_sec * 1000 + (now.tv_nsec + 500000) / 1000000);
+  return ((rc_clock_t)now.tv_sec * 1000 + ((rc_clock_t)now.tv_nsec + 500000) / 1000000);
```
hrydgard, line 222: Not sure I understand the +500000 addition here. There's not really any need for rounding here, in fact I think it's more consistent to round down always.

Jamiras: Agreed. For a monotonic timer the rounding is irrelevant.

PR #263, opened 2023-08-31
Better Swift support

MaddTheSane:
The module map makes it possible for Swift to find rcheevos headers not included in what it expects to be the umbrella header (`rcheevos.h`).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,0 +1,32 @@
+// swift-tools-version: 5.8
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "rcheevos",
+    products: [
+        // Products define the executables and libraries a package produces, and make them visible to other packages.
+        .library(
+            name: "rcheevos",
+            targets: ["rcheevos"]),
+    ],
+    dependencies: [
+        // Dependencies declare other packages that this package depends on.
+        // .package(url: /* package url */, from: "1.0.0"),
+    ],
+    targets: [
+        // Targets are the basic building blocks of a package. A target can define a module or a test suite.
+        // Targets can depend on other targets in this package, and on products in packages this package depends on.
+        .target(
+            name: "rcheevos",
+            dependencies: [],
+            path: ".",
+            exclude: ["src/rcheevos/rc_libretro.c"],
+            sources: ["include", "src/rcheevos", "src/rapi", "src/rhash"],
+            publicHeadersPath: "include",
+            cSettings: [
+                .define("RC_DISABLE_LUA")
+            ]),
+    ]
+)
--- a/include/module.modulemap
+++ b/include/module.modulemap
@@ -47,6 +47,10 @@ module rcheevos {
     
     module rc_runtime {
         header "rc_runtime.h"
+        export *
+    }
+    
+    module rc_runtime_types {
         header "rc_runtime_types.h"
         export *
     }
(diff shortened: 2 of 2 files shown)


include/module.modulemap
```diff
@@ -0,0 +1,66 @@
...
+  
+  module rc_api_runtime {
+	  header "rc_api_runtime.h"
+	  export *
+  }
+
+  module rc_api_user {
+	  header "rc_api_user.h"
+	  export *
+  }
```
Jamiras, line 27: I'm not really familiar with how this works, but from a quick Google, it seem like you'd want to group the "rc_api_" files under a single "rapi" module, and then everything else except "rc_hash" and "rc_url" in a "runtime" module.

Package.swift
```diff
@@ -0,0 +1,32 @@
...
+            name: "rcheevos",
+			dependencies: [],
+			path: ".",
+			exclude: ["src/rcheevos/rc_libretro.c"],
+			sources: ["include", "src/rcheevos", "src/rapi", "src/rhash"],
+			publicHeadersPath: "include",
+			cSettings: [
+				.define("RC_DISABLE_LUA")
+			]),
+    ]
```
Jamiras, line 31: All tabs should be converted to spaces. Same comment applies to other file.

Jamiras (commented):
I don't know how to test this. And as I won't be using it, it may drift out-of-sync.

Perhaps someone with more experience in the area can chime in.

I have left some feedback that I think should be addressed, but in the end this will probably end up being merged just with your word that it's beneficial.

include/module.modulemap
```diff
@@ -0,0 +1,67 @@
...
+            export *
+        }
+        
+        module rc_api_user {
+            header "rc_api_user.h"
+            export *
+        }
+
+        export *
+    }
```
Jamiras, line 31: I was thinking more along the lines of:
```
module rapi {
    header "rc_api_editor.h"
    header "rc_api_info.h"
    header "rc_api_request.h"
    header "rc_api_runtime.h"
    header "rc_api_user.h"
    export *
}
```

Or doesn't it work that way?

Is there some benefit to keeping them separate? If so, does the new intermediary module provide any benefits?

include/module.modulemap
```diff
@@ -0,0 +1,67 @@
...
+    module rc_error {
+        header "rc_error.h"
+        export *
+    }
+    
+    module rc_runtime {
+        header "rc_runtime.h"
+        header "rc_runtime_types.h"
+        export *
+    }
```
Jamiras, line 52: `rc_client` is definitely a runtime file. `rc_consoles` and `rc_error` are common files shared by the other modules. Their supporting code is with the other runtime files. I'm not sure if they warrant a "common" module, or should just be glopped in with the runtime files.

MaddTheSane: The reason I put `rc_runtime.h` and `rc_runtime_types.h` together is because I didn't know if the strange double-declaration of types would cause issues.

I'll just split them up as it doesn't seem to actually cause issues.

include/module.modulemap
MaddTheSane: You can put multiple headers into a single module, but the general rule of thumb (at least in Objective-C land) is one module per header.

Jamiras:
I'm not an expert, and no one else has provided any feedback. If you feel this is good-to-go, remove the Draft tag and I'll merge it.

PR #267, opened 2023-09-13
rc_client: Add functionality for feature parity with current DuckStation/PCSX2 implementations

stenzek:
This PR adds missing functionality to rc_client which is currently present in the DuckStation/PCSX2 achievements implementation, so that when I move them over (almost done), we don't lose any features.

[rc_client: add count enumeration values](https://github.com/RetroAchievements/rcheevos/commit/c766481f76df3c1abd86e5669846c7924541cf9c):
Enables clients to size arrays which handle incoming values of these enumeration types. If a new type is added, you might still end up dereferencing a null pointer, but that's better than a buffer/array overrun.

[rc_client: Add functions returning achievements/lbs/rich presence availability](https://github.com/RetroAchievements/rcheevos/commit/d7f2bdf68b85df020e3043f7b112077917a1f2e2) 
We're using these to determine which elements in the UI to display for the current game. For example, we don't want to poll Rich Presence and send it over to Discord if it's not available, because the fallback `Playing [gamename]` string is redundant when we're already reporting that information. We also use the presence of leaderboards to determine whether to display the breakout Achievements/Leaderboard menu, or go straight to the Achievement list.

[rc_client: Add leaderboard format mapping](https://github.com/RetroAchievements/rcheevos/commit/f3494ef4fc0da206d2ba73589ad04fab23c2bdff)
Based on discussion on Discord, to display a more context-specific title (e.g. "Time") in the leaderboard lists, rather than just "score", Jamiras suggested adding an enumeration of general value types, then mapping to that from the internal rcheevos format. This commit adds the mapping, which is present in each leaderboard's info. I intentionally did not add a "to_string" type function for converting it to a name, since this needs to be localized, and the most straight forward way of doing so is in the emulator. But if rc_client adds localization in the future, it's definitely something to consider.

[rc_client: Add scoreboard event](https://github.com/RetroAchievements/rcheevos/commit/e6d50f68f2ca6f4bb2507d3b0061fbcf94e88446)
Currently, rc_client ignores the response from the server after a leaderboard is submitted, containing the last score, best score, and current ranking. DuckStation and PCSX2 currently display this information on submit, therefore for future parity, we need to add a new event callback, because intercepting it on our side wouldn't be very clean or future-proof.

The event formats all scores in the same manner as the other `rc_client` structs. It does add another field to the main `rc_client_event_t` structure, if this is an issue, we could make it a union instead, and store a pointer to the achievement info in `rc_client_leaderboard_scoreboard_t`.

I wasn't 100% sure whether to include the top entries, as we don't currently use them, but somebody else might. The entries do get freed upon the callback returning though, which is in contrast to the other lists that rc_client creates. However, handing the memory off in the event would also mean that it would leak if the emulator didn't handle it, which certainly isn't desirable.


src/rcheevos/rc_client.c
```diff
@@ -1433,6 +1433,31 @@ static void rc_client_copy_achievements(rc_client_load_state_t* load_state,
...
+    case RC_FORMAT_SECONDS:
+    case RC_FORMAT_CENTISECS:
+    case RC_FORMAT_MINUTES:
+    case RC_FORMAT_SECONDS_AS_MINUTES:
+      return RC_CLIENT_LEADERBOARD_FORMAT_TIME;
+
+    case RC_FORMAT_SCORE:
+      return RC_CLIENT_LEADERBOARD_FORMAT_SCORE;
+
+    case RC_FORMAT_FRAMES:
```
Jamiras, line 1448: FRAMES should be a TIME format. It's implicitly divide by 60 and format as CENTISECS.

src/rcheevos/rc_client.c
```diff
@@ -2706,6 +2732,32 @@ void rc_client_destroy_achievement_list(rc_client_achievement_list_t* list)
...
+
+  subset = client->game->subsets;
+  result = 0;
+  for (; subset; subset = subset->next)
+  {
+    if (!subset->active)
+      continue;
+
+    result = 1;
+    break;
```
Jamiras, line 2753: these two lines should be in an `if (subset->public_.num_achievements > 0)`

src/rcheevos/rc_client.c
```diff
@@ -3242,6 +3294,32 @@ void rc_client_destroy_leaderboard_list(rc_client_leaderboard_list_t* list)
...
```
Jamiras, line 3315: these two lines should be in an if `(subset->public_.num_leaderboards > 0)`

src/rcheevos/rc_client.c
```diff
@@ -3345,6 +3423,62 @@ static void rc_client_submit_leaderboard_entry_retry(rc_client_scheduled_callbac
...
+    RC_CLIENT_LOG_ERR_FORMATTED(client, "Trying to raise scoreboard for unknown leaderboard %u", lboard_data->id);
+    return;
+  }
+
+  sboard.leaderboard_id = lboard_data->id;
+  rc_format_value(sboard.submitted_score, sizeof(sboard.submitted_score), response->submitted_score, leaderboard->format);
+  rc_format_value(sboard.best_score, sizeof(sboard.best_score), response->best_score, leaderboard->format);
+  sboard.new_rank = response->new_rank;
+  sboard.num_entries = response->num_entries;
+  sboard.top_entries = NULL;
```
Jamiras, line 3453: I prefer `memset`ing the entire structure to zero in case more fields are added later.

src/rcheevos/rc_client.c
```diff
@@ -3345,6 +3423,62 @@ static void rc_client_submit_leaderboard_entry_retry(rc_client_scheduled_callbac
...
+      response->num_top_entries, sizeof(rc_client_leaderboard_scoreboard_entry_t));
+    if (sboard.top_entries != NULL) {
+      unsigned i;
+      for (i = 0; i < response->num_top_entries; i++) {
+        sboard.top_entries[i].username = response->top_entries[i].username;
+        sboard.top_entries[i].rank = response->top_entries[i].rank;
+        rc_format_value(sboard.top_entries[i].score, sizeof(sboard.top_entries[i].score), response->top_entries[i].score,
+            leaderboard->format);
+      }
+      sboard.num_top_entries = response->num_top_entries;
```
Jamiras, line 3466: redundant with line 3454.

Jamiras:
> I wasn't 100% sure whether to include the top entries, as we don't currently use them, but somebody else might.

I'll add it later. The client DLL shows the top entries, and then the user's entry (if it's not in the top entries).
image

stenzek:
> I'll add it later. The client DLL shows the top entries, and then the user's entry (if it's not in the top entries).

I do include it now, but there's a behavioural difference versus the query leaderboard path, in that the pointers aren't valid after the event returns. That means that the emulator can't use just store it and render it in subsequent frames, like you can for normal queries.

Considering the leak point in the OP, I see two possible solutions:
 (1) have a "copy" function which takes the scoreboard structure, deep copies it, and have a corresponding free function.
 (2) persist the scoreboard in the internal leaderboard struct, overwriting the previous after submission.

(2) probably better matches the behaviour of the rest of rc_client, but neither really strike me as being great, unless you had another idea/approach?

Jamiras:
> There's a behavioural difference versus the query leaderboard path, in that the pointers aren't valid after the event returns. That means that the emulator can't use just store it and render it in subsequent frames, like you can for normal queries.

I think it's fine as the `server_error` subitem (for `RC_CLIENT_EVENT_SERVER_ERROR`) is also only valid during the callback. We should just document that behavior. I really don't see any benefit to keeping it around after the UI is done with it, which would be an unfortunate side-effect of (2).

And I'd argue that (1) isn't necessary either, as the client will most likely want to do some conversions on the data before showing it - such as converting the strings to wchar_t or storing the data in a structure that includes when a row should be highlighted.

stenzek:
Fair enough. I added a note to the structure declaration to hopefully avoid any chances of someone making the assumption.

include/rc_client.h
```diff
@@ -480,10 +505,46 @@ int rc_client_leaderboard_entry_get_user_image_url(const rc_client_leaderboard_e
...
+  /* The ID of the leaderboard which was submitted */
+  uint32_t leaderboard_id;
+  /* The value that was submitted */
+  char submitted_score[RC_CLIENT_LEADERBOARD_DISPLAY_SIZE];
+  /* The player's best submitted value */
+  char best_score[RC_CLIENT_LEADERBOARD_DISPLAY_SIZE];
+  /* The player's new rank within the leaderboard */
+  unsigned new_rank;
+  /* The total number of entries in the leaderboard */
+  unsigned num_entries;
```
Jamiras, line 531: I'm going to change these `unsigned`s to `uint32_t`s after I merge.

Jamiras (approved):

PR #268, opened 2023-09-16
Fix qsort with big achievement IDs or unlock times

CookiePLMonster:
As mentioned in https://github.com/RetroAchievements/rcheevos/commit/f207be8d01f7e39ced6a8d60303f872b30ee8bc3#r127529070, existing `qsort` predicates can under/overflow with input values big enough. Consider the following case for achievement IDs, but an equivalent scenario is possible for unlock times too.

* Achievement A has ID 5 and progress 50.0.
* Achievement B has ID 4294967295 (UINT_MAX) and progress 50.0.

As per the intended sorting rules, achievement A should come before achievement B. Howver, calculations done the way they are currently end up overflowing on `A->ID - B->ID`, in this case producing the result `6`. This, being a positive number, signals to `qsort` to put A after B.

This PR changes comparisons to be explicit and not rely on subtractions to avoid issues like this. A **new** test with such edge case data has been added and verified to fail without this PR and pass with it.

### DIFF
--- a/src/rcheevos/rc_client.c
+++ b/src/rcheevos/rc_client.c
@@ -2526,16 +2526,21 @@ static int rc_client_compare_achievement_unlock_times(const void* a, const void*
 {
   const rc_client_achievement_t* unlock_a = *(const rc_client_achievement_t**)a;
   const rc_client_achievement_t* unlock_b = *(const rc_client_achievement_t**)b;
-  return (int)(unlock_b->unlock_time - unlock_a->unlock_time);
+  if (unlock_b->unlock_time == unlock_a->unlock_time)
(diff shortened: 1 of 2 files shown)


Jamiras:
Ah, I see. But this is predicated on us having more than 2 billion achievements.

CookiePLMonster:
I agree this is super unlikely, but `unlock_time` may be far more likely to be affected - **if** trunctation is taking place (I don't know if it is), then it is potentially susceptible to a Y2038 issue.

Any modern compiles also makes this code branchless so there are really no downsides to doing it correctly.

Jamiras (approved):

PR #269, opened 2023-09-17
rc_client: Fix use-after-free when loading cached game

stenzek:
I hit this in PCSX2/DuckStation. Repro steps:

1. Launch an **unknown** game.
2. Shut down the emulator, unloading the game.
3. Launch the same game again.
4. The game load callback should fire immediately, passing the error.
5. Instead of returning null, `rc_client_begin_load_game()` returns the request which was free'd in `rc_client_begin_fetch_game_data() -> rc_client_load_error()`.
7. (in DS and PCSX2's case), abort the request, which writes to freed memory. It gets aborted because the freed pointer got stored, even though the callback had already fired (which is where I usually null it out), and on shutdown I cancel any outstanding requests.

Not sure if this is the best way to go about fixing it, I didn't look _too_ deep into these routines, but as best I can tell, we need to propagate the error back to `rc_client_load_game()`, so it doesn't return an already-freed `load_state`.

If you can think of a better way of fixing it, then I'll close the PR, but in the interim it serves as a potential fix and issue report :)

### DIFF
--- a/src/rcheevos/rc_client.c
+++ b/src/rcheevos/rc_client.c
@@ -64,7 +64,7 @@ typedef struct rc_client_load_state_t
   uint8_t hash_console_id;
 } rc_client_load_state_t;
 
-static void rc_client_begin_fetch_game_data(rc_client_load_state_t* callback_data);
+static int rc_client_begin_fetch_game_data(rc_client_load_state_t* callback_data);
 static void rc_client_hide_progress_tracker(rc_client_t* client, rc_client_game_info_t* game);
 static void rc_client_load_error(rc_client_load_state_t* load_state, int result, const char* error_message);
 static rc_client_async_handle_t* rc_client_load_game(rc_client_load_state_t* load_state, const char* hash, const char* file_path);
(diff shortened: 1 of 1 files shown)


src/rcheevos/rc_client.c
```diff
@@ -1939,7 +1939,8 @@ static rc_client_async_handle_t* rc_client_load_game(rc_client_load_state_t* loa
   else {
     RC_CLIENT_LOG_INFO_FORMATTED(client, "Identified game: %u (%s)", load_state->hash->game_id, load_state->hash->hash);
 
-    rc_client_begin_fetch_game_data(load_state);
+    if (!rc_client_begin_fetch_game_data(load_state))
+      return NULL;
```
Jamiras, line 1943: > If you can think of a better way of fixing it, then I'll close the PR

I think a better solution is to simply check to see if the load_state is still active before returning its handle on line 1945:

```
  return (client->state.load == load_state) ? &load_state->async_handle : NULL;
```

I'm working on writing a test to verify.

Jamiras: #270

stenzek: SGTM, I'll double check it with PCSX2 tonight.

stenzek: Yup, works fine.

PR #299, opened 2023-12-12
Misc fixes for rhash

CasualPokePlayer:
- Correct SEEK_END behavior for rc_file_seek_buffered_file. offset should be added, not substracted (the caller is supposed to send a negative offset here)
- Properly open UTF-8 paths on Windows (this requires conversion from UTF-8 to UTF-16 and using _wfopen_s/_wfopen).

### DIFF
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -7,6 +7,11 @@
 #include <stdio.h>
 #include <ctype.h>
 
+#if defined(_WIN32) && defined(_UNICODE)
+#define WIN32_LEAN_AND_MEAN
+#include <windows.h>
+#endif
+
 /* arbitrary limit to prevent allocating and hashing large files */
 #define MAX_BUFFER_SIZE 64 * 1024 * 1024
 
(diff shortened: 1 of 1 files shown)


src/rhash/hash.c
```diff
@@ -48,7 +53,34 @@ static struct rc_hash_filereader* filereader = NULL;
 
 static void* filereader_open(const char* path)
 {
+#if defined(_WIN32) && defined(_UNICODE)
```
Jamiras, line 56: I don't think we want the `&& defined(_UNICODE)`. The `UNICODE` and `_UNICODE` macros only specify what to map `TCHAR` and `tcslen` to. Since you're using the `_w` functions directly, this code should not be limited to a UNICODE compile. It could also work with an MBCS compile, or an unspecified compile.

CasualPokePlayer: The intent is to allow the support to not be compiled in, in case Windows versions without Unicode support (i.e. Win95/98/ME) care to be supported (granted, probably not really?), or perhaps a frontend is using this but also is not sending in UTF8 paths but rather ANSI paths (which granted, would arguably be wrong in the first place paths are specified to be utf8 in the filereader api, so maybe this isn't really a concern)

Jamiras: If that's the case, you should be using `WINVER` or something similar. I still argue that this functionality could/should work even if the bulk of the product is still using the `SetWindowTextA` line of APIs.

Jamiras (approved):

PR #300, opened 2023-12-14
Define and use macros for exporting public functions for dll/dylib/so usages, enforce cdcel for public functions/callbacks

CasualPokePlayer:


include/rc_export.h
```diff
@@ -0,0 +1,69 @@
...
+
+/* RC_SHARED should be defined when building rcheevos as a shared library (e.g. dll/dylib/so) */
+/* RC_STATIC should be defined when building rcheevos as a static library */
+/* BUILDING_RC should be defined when building rcheevos itself, and should not be defined for external code */
+
+/* TODO: BUILDING_RC currently only has an effect for RC_SHARED (this is abused for test code), perhaps when it is undefined internal definitions should be hidden? */
+
+#if !defined(RC_SHARED) && !defined(RC_STATIC)
+  #error RC_SHARED or RC_STATIC must be defined
+#endif
```
Jamiras, line 14: Rather than error out, I think it should default to RC_STATIC. You're probably the only consumer currently that doesn't just embed the c files directly in their executable, and not having a default would be a breaking change.

include/rc_export.h
```diff
@@ -0,0 +1,69 @@
...
+/* RC_C_LINKAGE typedef void (*rc_callback_t)(void) */
+
+#ifdef __cplusplus
+  #define RC_C_LINKAGE extern "C"
+#else
+  #define RC_C_LINKAGE
+#endif
+
+/* RC_CCONV should be used for public functions and callbacks, to enforce the cdecl calling convention, if applicable */
+/* RC_CCONV should be placed after the return type, and between the (* for callbacks */
```
Jamiras, line 32: Recommend adding an `and` to the second clause: 
```
and between the ( and * for callbacks */
```

include/rc_export.h
```diff
@@ -0,0 +1,69 @@
...
+/* RC_C_LINKAGE should be used for callbacks, to enforce the C calling convention */
+/* RC_C_LINKAGE should be placed before the return type, or before typedef if present */
+/* RC_C_LINKAGE void (*rc_callback)(void) */
+/* RC_C_LINKAGE typedef void (*rc_callback_t)(void) */
+
+#ifdef __cplusplus
+  #define RC_C_LINKAGE extern "C"
+#else
+  #define RC_C_LINKAGE
+#endif
```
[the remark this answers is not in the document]
Jamiras: This is also very verbose and prone to being forgotten. How is this better than doing the `extern "C" {` for the whole file?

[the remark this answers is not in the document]
CasualPokePlayer: I think my intent was to make the extern C usage explicit for function pointers rather than implicit, although I guess either way I did that anyways for RC_EXPORT, so it's probably not that much better in the first place (also extern "C" for function pointers is more academically correct if anything, in practice nearly every C++ implementation does not actually differ the calling convention depending on C / C++ linkage.

Linkage is the actual term used for extern "C" / extern "C++" anyways, and it makes sense given how it actually works. All extern "C" really does in practice is disable the name mangling (and possibly error for code that requires name mangling, e.g. overloads, templates), so symbols can be linked against C code, as it follows the same linkage as C code. It technically also enforces a C calling convention rather than a C++ calling convention, but for nearly every implementation these are the same.

include/rc_export.h
```diff
@@ -0,0 +1,69 @@
+#ifndef RC_EXPORT_H
+#define RC_EXPORT_H
+
+/* These macros control how callbacks and public functions are defined */
+
+/* RC_SHARED should be defined when building rcheevos as a shared library (e.g. dll/dylib/so) */
+/* RC_STATIC should be defined when building rcheevos as a static library */
+/* BUILDING_RC should be defined when building rcheevos itself, and should not be defined for external code */
+
+/* TODO: BUILDING_RC currently only has an effect for RC_SHARED (this is abused for test code), perhaps when it is undefined internal definitions should be hidden? */
```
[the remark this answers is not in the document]
CasualPokePlayer: That's probably better in practice, I'll do that.

include/rc_export.h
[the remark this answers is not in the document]
CasualPokePlayer: I suppose I could still use this idea to define RC_BEGIN_C_HEADER and RC_END_C_HEADER macros (or maybe some other name), which mght be slightly better than the contant #ifdef __cplusplus checks.

include/rc_export.h
```diff
@@ -0,0 +1,100 @@
...
+/* RC_CXX_GUARD_END should be placed after header declarations */
+
+/* example usage
+ *
+ * #ifndef RC_HEADER_H
+ * #define RC_HEADER_H
+ *
+ * #include <stdint.h>
+ *
+ * RC_CXX_GUARD_BEGIN
```
Jamiras, line 31: I'm still not sold on using #defines for this. Every library I've looked at just has the raw `#ifndef __cplusplus extern "C" {`

* https://github.com/madler/zlib/blob/643e17b7498d12ab8d15565662880579692f769d/zlib.h#L36-L38
* https://github.com/libsdl-org/SDL/blob/b937c54b665f134f241313eb45a4e0110807ae2c/include/SDL3/SDL_version.h#L34-L37
* https://github.com/openssl/openssl/blob/e4542332fa36eab6d6bbf33815bde433ade3b547/include/openssl/decoder.h#L25-L27

CasualPokePlayer: https://github.com/mgba-emu/mgba/blob/3a5642fcb81d3c279f8f385828e95ca3029d527d/include/mgba-util/common.h#L9-L15

Some libraries do use a macro for this (well, mGBA is the only one I know on hand)

CasualPokePlayer: RetroArch apparently also does it https://github.com/libretro/RetroArch/blob/2980eb7e1296498305b9c047c82050ffd4d85778/libretro-common/include/retro_common_api.h#L41-L59

Jamiras: > RetroArch apparently also does it 

Not always.
https://github.com/libretro/RetroArch/blob/2980eb7e1296498305b9c047c82050ffd4d85778/libretro-common/include/libretro.h#L30-L32

And definitely not consistently.
```
$ grep -rnI "extern \"C\" {" * | grep -v deps | wc -l
462
$ grep -rnI RETRO_BEGIN_DECLS * | grep -v deps | wc -l
209
```

Jamiras: The closest I can find to any sort of industry standard is [this document for writing gnu libraries](https://www.gnu.org/software/libtool/manual/html_node/C-header-files.html). It uses `BEGIN_C_DECLS` and `END_C_DECLS`, which do seem somewhat familiar, and googling for `BEGIN_C_DECLS` returns plenty of hits.

Would you be willing to change the defines to `RC_BEGIN_C_DECLS` and `RC_END_C_DECLS`?

CasualPokePlayer: I would be willing to do that. I mainly did CXX_GUARD as I only remember mGBA doing this and that's what it used (although in retrospect this is more an exception than the norm)

Jamiras (approved):

PR #319, opened 2024-02-25
runtime: add bounds checking to progress serialization

stenzek:
Currently, `rc_runtime_serialize_progress()` and `rc_runtime_deserialize_progress()` (and thus transitively their rc_client counterparts) do not perform any bounds checking on the provided data. Since save states are user-provided data, it should not be assumed that this data is well-formed, either from malicious intent, or random corruption.

This PR adds bounds checking to all buffer read/writes used for progress, and also a `serialized_size` output parameter to `rc_client_serialize_progress()`, which allows single-pass saving without first needing to compute the buffer size. A new error code, `RC_BUFFER_OVERFLOW` is returned when the provided buffer is of insufficient size, to differentiate from a general out-of-memory error, and/or corrupted data.

Usage of integers for offsets has also been replaced with `size_t` for consistency (rc_client used `size_t`), as well as the function signatures for `rc_runtime_serialize_progress()` and `rc_runtime_deserialize_progress()`.


Jamiras:
Closing for alternate solution in #321

PR #333, opened 2024-05-01
hash whole file for PSP homebrew files (EBOOT.PBP)

### LINKED ISSUE #244: PSP homebrew hashing
On the PSP, homebrew games (which often are ports/variations of games for other platforms) are quite popular and it seems people are quite excited to make achievements for them. Hashing them is a real problem though.

Homebrew games are extracted to a directory under /PSP/GAME on the memory stick, like say /PSP/GAME/DOOM. They contain an executable container called EBOOT.PBP which contains the game binary, its PARAM.SFO and icons and stuff, so effectively simply hashing this file might seem to be enough. Unfortunately homebrew games have more files than that which may be easiliy modified, and some are simply an engine running different data files, like the luaengine, so we really have to hash other files as well. But homebrew also often writes their own configuration and save data to the same directory, so simply hashing all files won't work reliably.

It might be necessary to do something like hash the eboot, then look up that hash in an official table to figure out what other files to hash as well to make a solid identification. I realize there's probably no infrastructure for that in retroachievements at this time, and it would be real tricky to maintain.

Of course, it's also a complication that there's often no "canonical" version of a particular homebrew game or engine, so achievement makers will just have to pick one. 

Anyway, I'm not sure how practical it is, but I think documenting the problem here is at least a start.

Jamiras:
closes #244

### DIFF
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -2770,6 +2770,14 @@ static int rc_hash_psp(char hash[33], const char* path)
   uint32_t size;
   md5_state_t md5;
 
+  /* https://www.psdevwiki.com/psp/PBP
+   * A PBP file is an archive containing the PARAM.SFO, primary executable, and a bunch of metadata.
+   * While we could extract the PARAM.SFO and primary executable to mimic the normal PSP hashing logic,
+   * it's easier to just hash the entire file. This also helps alleviate issues where the primary
+   * executable is just a game engine and the only differentiating data would be the metadata. */
+  if (rc_path_compare_extension(path, "pbp"))
+    return rc_hash_whole_file(hash, path);
+
   track_handle = rc_cd_open_track(path, 1);
   if (!track_handle)
     return rc_hash_error("Could not open track");
@@ -3785,6 +3793,10 @@ void rc_hash_initialize_iterator(struct rc_hash_iterator* iterator, const char* 
         if (rc_path_compare_extension(ext, "pce"))
         {
           iterator->consoles[0] = RC_CONSOLE_PC_ENGINE;
+        }
+        else if (rc_path_compare_extension(ext, "pbp"))
+        {
+          iterator->consoles[0] = RC_CONSOLE_PSP;
         }
         else if (rc_path_compare_extension(ext, "pgm"))
         {
--- a/test/rhash/test_hash.c
+++ b/test/rhash/test_hash.c
@@ -1832,6 +1832,37 @@ static void test_hash_psp_video()
   ASSERT_NUM_EQUALS(result_iterator, 0);
 }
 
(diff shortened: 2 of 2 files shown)


hrydgard:
This will accept some files that PPSSPP will not, such as EBOOTs with PSX isos inside (which then won't run anyway, so maybe it doesn't matter). Otherwise looks good, maybe it's fine as-is really.

Jamiras:
Is there some way to identify those sort of EBOOTs so we could exclude them from generating a hash? We already indirectly exclude UMD video discs because they don't have a `PSP_GAME` folder, where we expect to find the `EBOOT.BIN`.

Even if we don't explicitly exclude them from hashing, if the hash isn't registered on the server, it won't load achievements. And we have some policies around creating sets for the same game on alternate systems.

hrydgard:
Had to go look at our somewhat messy identification code for the details :)  [link 1](https://github.com/hrydgard/ppsspp/blob/master/Core/Loaders.cpp#L148) [link 2](https://github.com/hrydgard/ppsspp/blob/master/Core/Loaders.cpp#L202)

After first checking that the first four bytes are "\x00PBP", then read four bytes at offset 0x24 to get the offset of the internal archive, then read 4 bytes at that offset to get a magic number to look at.

This can be "NPUM" which means an embedded UMD (this is a game demo, most likely), or "PSIS" which means it's a PSX ISO. Otherwise, go ahead and hash.

For an extra check can also look at the PARAM.SFO file embedded in the PBP (which has a simple structure), if it's "CATEGORY" field is "ME", it's a PSX ISO.

TheFetishMachine:
> This can be "NPUM" which means an embedded UMD (this is a game demo, most likely), or "PSIS" which means it's a PSX ISO. Otherwise, go ahead and hash.

Just wanted to quickly mention that there's a usecase for wanting to hash PSP game demos as well; there's some that have exclusive content not present in the full game they're a demo of e.g. the Locoroco Halloween and Christmas demos, so I'd recommend not excluding them from hashing.

>For an extra check can also look at the PARAM.SFO file embedded in the PBP (which has a simple structure), if it's "CATEGORY" field is "ME", it's a PSX ISO.

Somewhat related, some people have expressed desire to be able to hash PSP PSX releases and link them to their respective PS1 entries and be able get achievements for them that way. I can see this making some sense (and there's some precedence with stuff like virtual console releases of games for Nintendo consoles getting linked), and while I'm not sure about it, there is always the possibility of a PSX rerelease having some exclusive and/or otherwise unavailable content. Might be worth considering as well.

Jamiras:
> Up to you whether this is all worth checking!

For now, I'm going to just let them hash and assume the hash won't resolve to a game on the server.

PR #334, opened 2024-05-04
Fix warning about hiding previous declaration in rc_client.c

AdmiralCurtiss:
Ran into this while building Dolphin. There's already a `scan` a few scopes up.

This block was also indented inconsistently for some reason, I've made it consistent with its surroundings.


Jamiras (commented):
Suggest also adding `-Wshadow` to the `SRC_CFLAGS` in Makefile so CI can catch this in the future.
https://github.com/RetroAchievements/rcheevos/blob/3decda081f98582e02d583474048dade799edb2a/test/Makefile#L110-L113

AdmiralCurtiss:
Sure.

Jamiras (approved):

PR #343, opened 2024-05-17
Add Accumulator Function for Remembering and Recalling a value.

redwizard42:
Adds a Remember condition to store a value into a designated accumulator for the current group (core, alt1, etc).  Defaults to accepting a single operand or can specify an operator and a second operand. At the beginning of an evaluation, the accumulator holds an unsigned integer with value 0.

Adds a Recall operand which evaluates to the value of the group's accumulator.

The Remember condition is a combining condition (similar to Add Source and Add Address) and does not break a chain.


redwizard42:
Example use case:

Access data at a calculated location at the end of a pointer.
Recalled value at the Add Address Line below evaluations to Pointer Value + (16-Bit Value at 0x1234) * 8 + 4.  The 8-bit value at this location is then compared to 0x1c.

```
Remember    Mem    16-Bit 0x1234 * Value        0x08
Remember    Recall               + Value        0x04
Remember    Recall               + Mem   32-Bit Pointer 
Add Address Recall
            Mem    8-Bit  0x00   = Value        0x1c
```

include/rc_runtime_types.h
```diff
@@ -106,7 +106,8 @@ enum {
   RC_OPERAND_LUA,            /* A Lua function that provides the value. */
   RC_OPERAND_PRIOR,          /* The last differing value at this address. */
   RC_OPERAND_BCD,            /* The BCD-decoded value of a live address in RAM. */
-  RC_OPERAND_INVERTED        /* The twos-complement value of a live address in RAM. */
+  RC_OPERAND_INVERTED,       /* The twos-complement value of a live address in RAM. */
+  RC_OPERAND_RECALL     /* The accumulator */
```
Jamiras, line 110: Please align the comment. Maybe expand on it too:
```
The value captured by the last RC_CONDITION_REMEMBER condition
```

src/rcheevos/condition.c
```diff
@@ -174,6 +174,7 @@ rc_condition_t* rc_parse_condition(const char** memaddr, rc_parse_state_t* parse
       case 'b': case 'B': self->type = RC_CONDITION_SUB_SOURCE; can_modify = 1; break;
       case 'c': case 'C': self->type = RC_CONDITION_ADD_HITS; break;
       case 'd': case 'D': self->type = RC_CONDITION_SUB_HITS; break;
+      case 'e': case 'E': self->type = RC_CONDITION_REMEMBER; can_modify = 1; break;
```
Jamiras, line 177: I generally try to assign letters that at least indirectly abbreviate the behavior. Given the available letters, I probably would have chosen `K` (for keep). Is there some reason you picked `E`?

src/rcheevos/rc_internal.h
```diff
@@ -89,12 +89,13 @@ typedef struct {
...
   lua_State* L;
 
-  rc_typed_value_t measured_value;  /* Measured */
-  uint8_t was_reset;                /* ResetIf triggered */
-  uint8_t has_hits;                 /* one of more hit counts is non-zero */
-  uint8_t primed;                   /* true if all non-Trigger conditions are true */
-  uint8_t measured_from_hits;       /* true if the measured_value came from a condition's hit count */
-  uint8_t was_cond_reset;           /* ResetNextIf triggered */
+  rc_typed_value_t measured_value;     /* Measured */
+  rc_typed_value_t accumulator_value;  /* Set Accumulator */
```
Jamiras, line 93: recommend `recall_value` or `remembered_value` to match other names.

src/rcheevos/operand.c
```diff
@@ -64,6 +65,54 @@ static int rc_parse_operand_lua(rc_operand_t* self, const char** memaddr, rc_par
...
+
+  if (i == 0)
+    return RC_INVALID_VARIABLE_NAME;
+
+  if (*aux != '}')
+    return RC_INVALID_VARIABLE_OPERAND;
+
+  ++aux;
+
+  if (strcmp(varName, "accumulator") == 0) {
```
Jamiras, line 105: Recommend renaming to `remembered` or `recall` to match other names

src/rcheevos/operand.c
```diff
@@ -64,6 +65,54 @@ static int rc_parse_operand_lua(rc_operand_t* self, const char** memaddr, rc_par
...
+    switch (*aux) {
+
+    default:
+      if (rc_is_valid_variable_character(*aux, i == 0)) {
+        varName[i] = *aux;
+        /* strncat_s(varName, RC_VALUE_MAX_NAME_LENGTH + 1, aux, 1); */
+        ++aux;
+      }
+      else {
+        ++aux;
```
Jamiras, line 83: It's unnecessary to modify a local variable before an explicit return.

Also, recommend inverting the logic to make it more straight-forward:
```
if (!rc_is_valid_variable_character(*aux, i == 0))
  return RC_INVALID_VARIABLE_NAME;

varName[i] = *aux++;
break;
```

src/rcheevos/operand.c
```diff
@@ -64,6 +65,54 @@ static int rc_parse_operand_lua(rc_operand_t* self, const char** memaddr, rc_par
...
+        return RC_INVALID_VARIABLE_NAME;
+      }
+      break;
+
+    case '\0':/* end of string */
+    case '_': /* next condition */
+    case 'S': /* next condset */
+    case ')': /* end of macro */
+    case '$': /* maximum of values */
+      return RC_INVALID_VARIABLE_OPERAND;
```
Jamiras, line 93: None of these should matter if we're expecting the variable name to be surrounded by brackets. Either it'll be a valid character (part of the name) or it won't.

src/rcheevos/operand.c
```diff
@@ -64,6 +65,54 @@ static int rc_parse_operand_lua(rc_operand_t* self, const char** memaddr, rc_par
...
+    }
+  }
+
+  if (i == 0)
+    return RC_INVALID_VARIABLE_NAME;
+
+  if (*aux != '}')
+    return RC_INVALID_VARIABLE_OPERAND;
```
Jamiras, line 101: This should also just be `RC_INVALID_VARIABLE_NAME`. Either we found a valid name enclosed by brackets, or we didn't.

src/rcheevos/operand.c
```diff
@@ -160,7 +209,7 @@ int rc_parse_operand(rc_operand_t* self, const char** memaddr, uint8_t is_indire
       }
       allow_decimal = 1;
       /* fall through */
-    case 'v': case 'V': /* signed integer constant */
+    case 'v': case 'V': /* signed integer constant or variable */
```
Jamiras, line 212: How does this support `variable`?

src/rcheevos/operand.c
```diff
@@ -231,6 +280,13 @@ int rc_parse_operand(rc_operand_t* self, const char** memaddr, uint8_t is_indire
           self->value.num = (unsigned)value;
       }
       break;
+    case '{': /* variable or accumulator */
```
Jamiras, line 283: I don't think it's necessary to call out the accumulator here, it's just a special variable.

src/rcheevos/trigger.c
```diff
@@ -176,6 +178,9 @@ int rc_evaluate_trigger(rc_trigger_t* self, rc_peek_t peek, void* ud, lua_State*
     char sub_primed = 0;
 
     do {
+      /* Clear the accumulator before each condset; matches behavior of values */
+      eval_state.accumulator_value.type = RC_VALUE_TYPE_UNSIGNED;
+      eval_state.accumulator_value.value.i32 = 0;
```
Jamiras, line 183: This should be in `rc_test_condset_internal` where it's clearing out the `add_value`.

test/rcheevos/test_trigger.c
```diff
@@ -1940,6 +1940,121 @@ static void test_large_memref_not_shared() {
...
+  ASSERT_NUM_EQUALS(trigger->measured_target, 4U);
+
+  /* condition is true - previously met */
+  assert_evaluate_trigger(trigger, &memory, 1);
+  assert_hit_count(trigger, 0, 2, 3U);
+  assert_hit_count(trigger, 0, 4, 3U);
+  ASSERT_NUM_EQUALS(trigger->measured_value, 6U);
+  ASSERT_NUM_EQUALS(trigger->measured_target, 4U);
+}
+
```
Jamiras, line 2057: I would also like to see the following tests:
* Something chaining the accumulator (remember accumulator * value)
* Something using multiple accumulators in the same group (remember, do something, remember something else, do something else)
* Something using multiple accumulators in the same group where one is used by a Pause condition and one is used by a normal condition. (remember, pause, remember something else, do something else)
  - also pause second (remember something else, do something else, remember, pause)
* Something using the accumulator without setting it first.

Jamiras: I see some of these might be covered in the value tests below.

src/rcheevos/condition.c
redwizard42: Similarity to add/sub source and add/sub hits in general use case, so I just followed to the next letter. Is K your reference still? Easy enough to switch if so.

src/rcheevos/operand.c
redwizard42: I think I added the ++aux there it consume the } before returning. I'll review further once I have time to sit down and look.

src/rcheevos/operand.c
redwizard42: Stale comment from noodling around before settling on { } and I missed it during my comment cleanup passes.

test/rcheevos/test_trigger.c
redwizard42: Yeah, some cases were functionally tested in the value tests which I added first. For those that are should I  include coverage in trigger as well? May be some merit. I'll review to add uncovered cases next time I work on this.

src/rcheevos/trigger.c
redwizard42: At current time, keeping it where it is I should allow the non-pause pass to use the last remembered value from the pause pass.  This could have some logic-readability benefit.

Example:
https://retroachievements.org/game/29669
If I needed a pause in an achievement, I could set up the bulk of the very long pointer chain in the pause, then re-use it in the non-pause pass.  This is effectively a side-effect though, so perhaps avoiding a side effect is better than having some logic/readbility convenience.

Moving it to the internal test would effectively many pause pass and non-pause pass have a "different" recall accumulator, which is perhaps more consistent behavior.

Thoughts?

test/rcheevos/test_trigger.c
redwizard42: I think I have coverage for all of these in either test_trigger.c or test_value.c now. If I missed one, let me know.

src/rcheevos/condition.c
redwizard42: Updated it to K.

src/rcheevos/trigger.c
Jamiras: I agree that sharing it between the pause and non-pause passes makes sense. I still think this is too far up. You've ended up putting this logic in three places where `rc_test_condset` is called. (There's a fourth in the unit tests that you missed). It should just be _in_ `rc_test_condset` (after the empty check).

redwizard42: That makes send to me. I've moved it.

src/rcheevos/condset.c
```diff
@@ -417,6 +427,10 @@ int rc_test_condset(rc_condset_t* self, rc_eval_state_t* eval_state) {
     return 1;
   }
 
+  /* initialize recall value so each condition set has a functionally new recall accumulator */
+  eval_state->recall_value.type = RC_VALUE_TYPE_UNSIGNED;
+  eval_state->recall_value.value.i32 = 0;
```
Jamiras, line 432: You should set `u32` for `UNSIGNED`.

src/rcheevos/operand.c
```diff
@@ -64,6 +65,37 @@ static int rc_parse_operand_lua(rc_operand_t* self, const char** memaddr, rc_par
   return RC_OK;
 }
 
+static int rc_parse_operand_variable(rc_operand_t* self, const char** memaddr/*, rc_parse_state_t* parse*/) {
```
Jamiras, line 68: Please remove the commented-out `parse`  variable

src/rcheevos/operand.c
```diff
@@ -231,6 +263,13 @@ int rc_parse_operand(rc_operand_t* self, const char** memaddr, uint8_t is_indire
           self->value.num = (unsigned)value;
       }
       break;
+    case '{': /* variable */
+      ++aux;
+      ret = rc_parse_operand_variable(self, &aux/*, parse*/);
```
Jamiras, line 268: Please remove the unused `parse` variable.

src/rcheevos/trigger.c
```diff
@@ -176,6 +176,7 @@ int rc_evaluate_trigger(rc_trigger_t* self, rc_peek_t peek, void* ud, lua_State*
     char sub_primed = 0;
 
     do {
+      /* Clear the accumulator before each condset; matches behavior of values */
```
Jamiras, line 179: Comment is no longer necessary here.

test/rcheevos/test_trigger.c
```diff
@@ -1940,6 +1940,172 @@ static void test_large_memref_not_shared() {
...
+  memory_t memory;
+  rc_trigger_t* trigger;
+  char buffer[512];
+
+  memory.ram = ram;
+  memory.size = sizeof(ram);
+
+  assert_parse_trigger(&trigger, buffer, "K:0xH00_{recall}<3.4._K:0xH00*2_{recall}>0xH01_K:0xH00*2_P:{recall}=2");
+
+/* pause checks 0*2=2, not paused. Condition 2 gets hit since recalled value < 3 */
```
Jamiras, line 2069: indent is inconsistent.

src/rcheevos/condset.c
redwizard42: Good catch.

redwizard42:
Comments and indentation should be cleaned up now.

PopovMark:
Is there only one slot for the accumulator or will you do multiple, kinda like variables? From what I can tell you can only store one value at any given time, why not make it 4 or 8 values?

PR #358, opened 2024-07-25
add rc_client_set_filereader and rc_client_get_default_cdreader

Jamiras:
Allows a client to register a custom filereader/cdreader for calculating hashes when calling `rc_client_begin_identify_and_load_game`.

Replaces direct calls to `rc_hash_init_custom_filereader` and `rc_hash_init_custom_cdreader`. Also supports passing the readers through the external interface to the DLL.

https://discord.com/channels/476211979464343552/757767535293890682/1265528322943881306


Jamiras:
I think I've come up with an alternate solution that's much cleaner. Going to switch this to draft while I experiment.

PR #366, opened 2024-09-02
Add DTCM to DS/DSi memory maps

### LINKED ISSUE #291: Add DTCM to DS/DSi Memory Maps
https://problemkaputt.de/gbatek.htm#dsmemorycontrolcacheandtcm
https://problemkaputt.de/gbatek.htm#armcp15tightlycoupledmemorytcm

TCM is high speed memory on the ARM9 CPU itself. There are two types of TCM, one meant for Instructions (ITCM) and one meant for Data (DTCM). ITCM is probably unlikely to be useful for achievements, but DTCM seems to be useful. Okamiden for the DS appears to store all important savefile flags in this region rather than in Main RAM.

There is a question however where exactly should it be put. Should it be placed right after the Main RAM for each console (which would result in a odd memory map when comparing the two, and could be potentially problematic if you want to support running DSi enhanced games in either DS or DSi mode), or should it be placed after what would be the Main RAM for DSi, and leave a 12MiB hole for DS?

CasualPokePlayer:
Resolves #291


src/rcheevos/consoleinfo.c
```diff
@@ -721,16 +721,29 @@ static const rc_memory_regions_t rc_memory_regions_n64 = { _rc_memory_regions_n6
 /* ===== Nintendo DS ===== */
 /* https://www.akkit.org/info/gbatek.htm#dsmemorymaps */
 static const rc_memory_region_t _rc_memory_regions_nintendo_ds[] = {
-    { 0x000000U, 0x3FFFFFU, 0x02000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "System RAM" }
+    { 0x0000000U, 0x03FFFFFU, 0x02000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "System RAM" },
+    /* To keep DS/DSi memory maps aligned, padding is set here for the DSi's extra RAM */
+    { 0x0400000U, 0x0FFFFFFU, 0x02400000U, RC_MEMORY_TYPE_UNUSED, "Unused (DSi exclusive)" },
```
Jamiras, line 726: Why is this padding necessary? We have a separate map for DSi. 

Is there concern that a DS game will be loaded in DSi mode? As long as the game is associated to DS in the database, it should use the DS map over the DSi map, even if the emulator is running in DSi mode. In that case, the "real address" of 0xE0000000 becomes the thing that will line up the data correctly.

I guess this is a simplification so an emulator running in DSi mode doesn't have to know it's a DS game that's loaded? I'm weary of leaving a 12MB gap here as the dev tools will still allocate a 16MB buffer when searching, and developers will wonder about the large gap when building achievements.

CasualPokePlayer: It could be something to consider for DSi *enhanced* games, which might have reason for the user to use DS or DSi, similar to GB/C. (especially since DS is a lot easier to setup in an emulator compared to DSi).

Jamiras: https://github.com/RetroAchievements/RAIntegration/pull/970 allows DS games to be loaded in a DSi context, which implies the DSi memory map. As such, the DTCM will be expected to be located at $1000000. If that same game is loaded in a DS context, I guess we should ensure the DTCM is also located at $1000000 so the achievement logic doesn't have to check both.

If the game is running in DSi mode, shouldn't the full 16MB of system memory be exposed? Wouldn't the DSi enhancements also use that? What is the probability of the important memory for achievements being shifted around as the result of loading a DSi-enhanced game in a DSi context instead of a DS context?

Jamiras (approved):

src/rcheevos/consoleinfo.c
CasualPokePlayer: The full 16 MiBs probably should be exposed (DSi enhanced games have access to such, unlike non-DSi enhanced games). Main gameplay probably wouldn't have any important memory in the latter memory for DSi enhanced games (outside of some heap memory which you'd be following a pointer anyways).

PR #368, opened 2024-09-03
add another check for game being unloaded before calling load callback

Jamiras:
https://discord.com/channels/476211979464343552/757767535293890682/1280531659812507732

### DIFF
--- a/src/rc_client.c
+++ b/src/rc_client.c
@@ -1455,13 +1455,16 @@ static void rc_client_activate_game(rc_client_load_state_t* load_state, rc_api_s
           start_session_response->num_unlocks, RC_CLIENT_ACHIEVEMENT_UNLOCKED_SOFTCORE);
     }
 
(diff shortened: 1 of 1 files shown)


CasualPokePlayer:
I'm not quite satisfied by this. This doesn't solve the underlying race condition occurring. `rc_client_free_game` needs to occur while there is no chance for another thread to interact with the client's `rc_client_game_info_t`. Or more frankly, you should not be reading `client->game` outside of the `client->state.mutex` lock. That, or you need to ensure `client->game` wouldn't suddenly be set to NULL and free'd while another thread is using it via some other mechanism.

Jamiras:
Made another pass. The `client->state.load` object is no longer detached at the start of `rc_client_activate_game`. If `client->state.load->game` is `client->game`, `rc_client_unload_game` will no longer free it, allowing `rc_client_activate_game` to do so after calling its callback.

PR #370, opened 2024-09-08
Add hashing for ZX Spectrum

CasualPokePlayer:
Just do full file hashes (going off all supported extensions in BizHawk: *.tzx, *.tap, *.pzx, *.csw, *.wav, *.dsk)

### DIFF
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -3978,6 +3978,10 @@ void rc_hash_initialize_iterator(struct rc_hash_iterator* iterator, const char* 
         {
           iterator->consoles[0] = RC_CONSOLE_THOMSONTO8; /* disk */
         }
+        else if (rc_path_compare_extension(ext, "scl"))
+        {
+          iterator->consoles[0] = RC_CONSOLE_ZX_SPECTRUM;
+        }
         break;
 
       case 't':
@@ -3994,7 +3998,8 @@ void rc_hash_initialize_iterator(struct rc_hash_iterator* iterator, const char* 
         {
           iterator->consoles[0] = RC_CONSOLE_ELEKTOR_TV_GAMES_COMPUTER;
         }
-        else if (rc_path_compare_extension(ext, "tzx"))
+        else if (rc_path_compare_extension(ext, "trd") ||
+                 rc_path_compare_extension(ext, "tzx"))
         {
           iterator->consoles[0] = RC_CONSOLE_ZX_SPECTRUM;
         }
@@ -4030,10 +4035,6 @@ void rc_hash_initialize_iterator(struct rc_hash_iterator* iterator, const char* 
         else if (rc_path_compare_extension(ext, "woz"))
         {
           iterator->consoles[0] = RC_CONSOLE_APPLE_II;
-        }
-        else if (rc_path_compare_extension(ext, "wav"))
-        {
-          iterator->consoles[0] = RC_CONSOLE_ZX_SPECTRUM;
         }
         break;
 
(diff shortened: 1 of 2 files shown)


src/rhash/hash.c
```diff
@@ -4015,6 +4031,10 @@ void rc_hash_initialize_iterator(struct rc_hash_iterator* iterator, const char*
         {
           iterator->consoles[0] = RC_CONSOLE_APPLE_II;
         }
+        else if (rc_path_compare_extension(ext, "wav"))
```
Jamiras, line 4034: I'm wary that `.wav` files wouldn't be exactly identical for multiple users.

src/rhash/hash.c
```diff
@@ -3982,6 +3994,10 @@ void rc_hash_initialize_iterator(struct rc_hash_iterator* iterator, const char*
         {
           iterator->consoles[0] = RC_CONSOLE_ELEKTOR_TV_GAMES_COMPUTER;
         }
+        else if (rc_path_compare_extension(ext, "tzx"))
+        {
+          iterator->consoles[0] = RC_CONSOLE_ZX_SPECTRUM;
```
Jamiras, line 3999: Looking at some of the "official" collections, most files seem to be `tzx`, `tap`, `z80`, and `scl`. You're not handling `z80` or `scl`. Should we? It looks like `z80` is a memory snapshot, which might be convenient for faster boots, but completely eliminates the possibility that it was dumped from a user's personal collection. I think we should probably ignore it. `scl` appears to be a version of `trd` that ignores unused sectors, so it should still provide a consistent dump.

The [libretro core](https://docs.libretro.com/library/fuse/#extensions) does support `scl`. It also supports `rzx` and `trd`. `rzx` appears to be a TAS type file, so we shouldn't consider it, but `trd` appears to be another disk format.

Are `csw` and `psx` uncommon or deprecated? They're unique enough that I'd be fine leaving them in, but then we should also consider adding `trd`.

| extension | bizhawk | fuse | include? |
| --- | ------- | ---- | -------- |
| csw | yes | no | maybe? |
| dsk | yes | no | probably |
| pzx | yes | no | maybe? |
| rzs | no | yes | no |
| scl | no | yes | maybe? |
| tap | yes | yes | yes |
| trd | no | yes |  maybe? |
| tzx | yes | yes | yes |
| wav | yes | no | maybe not? |
| z80 | no | yes | no |

src/rhash/hash.c
```diff
@@ -3427,6 +3428,7 @@ int rc_hash_generate_from_file(char hash[33], uint32_t console_id, const char* p
     case RC_CONSOLE_VIRTUAL_BOY:
     case RC_CONSOLE_WASM4:
     case RC_CONSOLE_WONDERSWAN:
+    case RC_CONSOLE_ZX_SPECTRUM:
```
Jamiras, line 3431: Please add unit tests for at least `tap` and `tzx`. You can copy the Apple II or Commodore tests. Since Spectrum is using full file hash, it wouldn't be difficult to enumerate all supported extensions.

src/rhash/hash.c
CasualPokePlayer: It's apparently just used to be interpreted as a tape. I'm not sure how common dumps are actually using a .wav file here.

src/rhash/hash.c
CasualPokePlayer: The BizHawk core doesn't handle z80 or scl. It might be desirable to include these anyways (fwiw, BizHawk doesn't use the iterator, as it already identifies the file at this point, this is just for convenience libretro)

src/rhash/hash.c
hrydgard: Presumably different recordings of old tapes would result in different .wav files, so this only makes sense if there's a "canonical" set of .wav recordings .. otherwise, maybe best not to support .wav, since there are other formats that might make more sense?

Jamiras (approved):

PR #375, opened 2024-10-07
Expand SNES memory exposure

CasualPokePlayer:
Alt approach against #371 

1. Expand Cartridge RAM from 128KiB to 512KiB. This is the maximum theoretically mappable for any official ROM types, possibly used in homebrews. This limit is also what Snes9x and other emulators internally limit Cartridge RAM to (although notably, libretro Snes9x limits Cartridge RAM to 128KiB for purposes of exposing memory, leaving the rest present but unexposed. This seems to be an artifact of the old 128KiB limit older Snes9x imposed, probably a trivial fix on that side anyways).
2. Set the "native address" of Cartridge RAM to 0x1000000, outside of the native addressing space. The old value used is only necessarily valid for LoROM; HiROM and SA-1 (BW-RAM) have Cartridge RAM elsewhere, and even then, the old value would only work assuming > 128KiB of Cartridge RAM wasn't mapped. Nearly all libretro cores do not implement explicit memory maps in the first place, the only 2 which do are bsnes-mercury and Snes9x 2010, which are known to be problematic (and thus unsupported) already (probably due to this quirk with Cartridge RAM if anything).
3. Map SA-1 I-RAM after Cartridge RAM. Unlike BW-RAM, this isn't SaveRAM (and thus not exposed as such by emulators), so it's a separate memory block. It is mapped outside of the native addressing space as its native location is typically just SNES WRAM for other cartridge types.

This approach has already been implemented in BizHawk (for all Snes9x, Faust, BSNES, and BSNESv115+ cores), so it could be tested against such (with its latest dev builds).


Jamiras (approved):
I think this implementation makes sense based on the comments [here](https://github.com/RetroAchievements/rcheevos/pull/371#discussion_r1782014121).

wolfman2000:
I'm good with that.

PR #376, opened 2024-10-11
Use better Win32 mutexes for rc_mutex_t

CasualPokePlayer:
https://learn.microsoft.com/en-us/archive/msdn-magazine/2012/november/windows-with-c-the-evolution-of-synchronization-in-windows-and-c

On Windows, a simple Mutex HANDLE (using *Mutex functions) seems like an appropriate implementation for rc_mutex_t. While it does work, it is horribly inefficient and actually overkill, as on Win32, a Mutex is a synchronization object capable of inter-process locking, thus requiring entering kernel code always when using a *Mutex function (going into kernel code implicitly being fairly expensive).

A critical section works similarly, except it is not capable of inter-process locking. This is much better for rc_mutex_t, which is obviously not expecting to be used in a different process. Critical sections are also capable of spinlooping for a little bit of time before trying to acquire the critical section, thus potentially avoiding an expensive system call when contention goes away quickly.

Windows Vista introduced slim reader/writer (SRW) locks. These are very lightweight, mostly implemented in user mode (only going to kernel if it decides it should just outright sleep), and using newer (more efficient) "keyed event objects" (which only have 1 actual kernel object allocated for all objects, only used as a last resort for critical sections due to backwards compatibility reasons). They aren't a direct replacement for critical sections as they are incapable of recursive acquires, but that's easy enough to implement by tracking the current thread owner.

In case code is compiled with Vista or later as the minimum version (WINVER >= 0x0600) then SRW locks will be used, otherwise critical sections will be used.

This commit also does a minor fix in rc_compat.h with libogc mutexes, as previously it was just including pthread.h and typedef'ing rc_mutex_t against pthread_mutex_t. Not sure why this ever compiled exactly (does that indicate GC/Wii can just use pthread mutexes? Or was RC_NO_THREADS actually being defined?), but regardless I've changed it to properly typedef against libogc's mutex_t.

### DIFF
--- a/src/rc_compat.c
+++ b/src/rc_compat.c
@@ -59,7 +59,7 @@ int rc_snprintf(char* buffer, size_t size, const char* format, ...)
 
   va_start(args, format);
 
-#ifdef __STDC_WANT_SECURE_LIB__
+#ifdef __STDC_SECURE_LIB__
   result = vsprintf_s(buffer, size, format, args);
 #else
   /* assume buffer is large enough and ignore size */
@@ -74,7 +74,7 @@ int rc_snprintf(char* buffer, size_t size, const char* format, ...)
 
 #endif
 
-#ifndef __STDC_WANT_SECURE_LIB__
+#ifndef __STDC_SECURE_LIB__
 
 struct tm* rc_gmtime_s(struct tm* buf, const time_t* timer)
 {
--- a/src/rc_compat.c
+++ b/src/rc_compat.c
@@ -106,6 +106,7 @@ void rc_mutex_init(rc_mutex_t* mutex)
 void rc_mutex_destroy(rc_mutex_t* mutex)
 {
   /* Nothing to do here */
+  (void)mutex;
 }
 
 void rc_mutex_lock(rc_mutex_t* mutex)
--- a/src/rc_compat.c
+++ b/src/rc_compat.c
@@ -1,5 +1,6 @@
 #include "rc_compat.h"
 
+#include <assert.h>
 #include <ctype.h>
 #include <stdarg.h>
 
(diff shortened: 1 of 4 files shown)


src/rc_compat.c
```diff
@@ -88,32 +89,78 @@ struct tm* rc_gmtime_s(struct tm* buf, const time_t* timer)
...
+    assert(mutex->owner == 0 && mutex->count == 0);
+    mutex->owner = current_thread;
+    mutex->count = 1;
+  }
+}
+
+void rc_mutex_unlock(rc_mutex_t* mutex)
+{
+  if (mutex->owner == GetCurrentThreadId()) {
+    if (--mutex->count == 0) {
```
Jamiras, line 130: Recommend having an `assert(mutex->count > 0)` here too. It actually makes more sense than having one when acquiring the lock.

src/rc_compat.h
```diff
@@ -1,6 +1,11 @@
 #ifndef RC_COMPAT_H
 #define RC_COMPAT_H
 
+#ifdef _WIN32
+ #define WIN32_LEAN_AND_MEAN
```
Jamiras, line 5: This needs to be wrapped in an `#ifndef`:
```
2>E:\Source\RetroAchievements\RAIntegration\rcheevos\src\rc_compat.h(5,10): warning C4005: 'WIN32_LEAN_AND_MEAN': macro redefinition
```

src/rc_compat.c
```diff
@@ -88,32 +89,78 @@ struct tm* rc_gmtime_s(struct tm* buf, const time_t* timer)
...
+  /* Nothing to do here */
+  (void)mutex;
+}
+
+void rc_mutex_lock(rc_mutex_t* mutex)
+{
+  DWORD current_thread = GetCurrentThreadId();
+  if (mutex->owner == current_thread) {
+    ++mutex->count;
+    assert(mutex->count > 0);
```
Jamiras, line 117: Missing include.
```
1>E:\Source\RetroAchievements\RAIntegration\rcheevos\src\rc_compat.c(116,5): warning C4013: 'assert' undefined; assuming extern returning int
```

src/rc_compat.h
```diff
@@ -77,10 +82,22 @@ RC_BEGIN_C_DECLS
...
+     DWORD owner;
+     DWORD count;
+   #else
+     /* Pre-Vista must use a critical section */
+     CRITICAL_SECTION critical_section;
+   #endif
+   } rc_mutex_t;
+ #elif defined(GEKKO)
+  #include <ogcsys.h>
+  typedef mutex_t rc_mutex_t;
```
Jamiras, line 100: The GEKKO code was written by another user. I have no way to test it.

See #312.

src/rc_compat.c
CasualPokePlayer: That's odd, there's an `#include <assert.h>` at the top of the file, which should include assert just fine?

Jamiras: Yep. I might have screwed something up copying the changes across.

src/rc_compat.h
CasualPokePlayer: Looking more deeply at the libogc docs indicates the usage is still wrong anyways: https://libogc.devkitpro.org/mutex_8h.html

Nearly all the functions take the mutex_t by value rather than by reference (only init takes it by reference). The LWP_MutexInit call also seems to be nonsense (passing NULL for the bool use_recursive seems unintended).

I could move fixes for that in a separate PR (presumingly some investigation might by warranted by RetroArch's side, RC_NO_THREADS might be getting defined hence why that is seemingly building at all).

CasualPokePlayer: Seems RC_NO_THREADS is actually getting defined on RetroArch's side: https://github.com/libretro/RetroArch/blob/228426880babb529fae8f01598a6d34698ad3b18/griffin/griffin.c#L194

Jamiras (approved):

PR #377, opened 2024-10-12
Do fixes for Gekko mutexes, implement 3DS mutexes

CasualPokePlayer:
These lines from RetroArch probably should be removed after this PR is in RetroArch's rcheevos copy. https://github.com/libretro/RetroArch/blob/228426880babb529fae8f01598a6d34698ad3b18/griffin/griffin.c#L193-L195

### DIFF
--- a/src/rc_compat.c
+++ b/src/rc_compat.c
@@ -177,22 +177,47 @@ void rc_mutex_unlock(rc_mutex_t* mutex)
 
 void rc_mutex_init(rc_mutex_t* mutex)
 {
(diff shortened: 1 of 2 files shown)


Jamiras (approved):
I have no way to test this. Looks reasonable.

Jamiras:
https://github.com/RetroAchievements/rcheevos/pull/378#issuecomment-2415564923
> Note probably want to introduce an !defined(_3DS) for the _XOPEN_SOURCE define guard when https://github.com/RetroAchievements/rcheevos/pull/377 is merged.

PR #378, opened 2024-10-12
Define pthread mutex as recursive

CasualPokePlayer:
For consistent semantics against other rc_mutex_t implementations (mostly more making sense with my other PRs; regardless, the current Windows mutex implementation already has recursive semantics).


Jamiras (approved):
I don't think the code is currently reliant on recursive locks. There's a few places with explicit comments indicating a lock is assumed to be held prior to calling the function (i.e. https://github.com/RetroAchievements/rcheevos/blob/32917bdddf4982e62047862c6633e7671aaaf2cb/src/rc_client.c#L5005) to eliminate the need for recursive locks.

Having said that, the changes seem reasonable, and do provide a better level of compatibility in case recursive locks end up getting introduced.

CasualPokePlayer:
Note probably want to introduce an `!defined(_3DS)` for the `_XOPEN_SOURCE` define when https://github.com/RetroAchievements/rcheevos/pull/377 is merged.

PR #386, opened 2024-11-16
Optimize CIA hashing

CasualPokePlayer:
The section of data between the header and ExeFS normally has to be "decrypted" to obtain the right AES state. However, we can just abuse the way CBC works and get the right IV regardless, by just reading the previous encrypted block before the ExeFS.


Jamiras (approved):
Verified this generates the same hashes for the handful of .cia files I have.

PR #408, opened 2025-03-23
Remove HAVE_LUA define

Jamiras:
After seven years of carrying around the code but never enabling it, I've finally made the decision to remove it as much as possible without breaking any of the API signatures.

LUA was originally seen as a potential solution for dealing with pointers and reusing the results of complex calculations. The first is now possible via AddAddress, and the second via Remember/Recall.

There is still other functionality that LUA could provide, like scanning blocks of memory and traversing linked lists, but we're very hesitant to support any sort of iteration construct as the developer could easily write an infinite loop (or just an inefficient one) and crash the emulator or introduce so much overhead to achievement processing that the gameplay would be affected. When we introduce functionality to do these things, we'll can add safeguards to limit the number of iterations (similar to safeguard we added where dividing by zero results in 0 instead of a crash).

To this end, I've removed all internal references to LUA. LUA was never exposed through the `rc_client` interface, so there should be very few implementations even referencing the external functions that were affected. Even so, I opted to renamed the parameters in the affected functions to `void* unused_L` and `int unused_funcs_idx` to obscure the legacy (never enabled) functionality.


Souzooka:
If I can say my piece, it's a bit sad to see this flag go. There's a bit of a silver lining to this though and something I hope which is considered when/if Lua is ever considered as a tool again, and that is that this current implementation was doomed to cause tons of technical debt if it was ever utilized. This is because it basically required each script to have its own Lua script if they used the flag, which would require duplicated code for every single asset if impossible calculations in rcheevos were necessary.

While math is somewhat improved due to remember/recall, it's still somewhat limited due to the lack of additional intermediary variables (which can manifest itself in pointer arithmetic, but a very popular example would be for example calculating 2d dist2, which would require 2 remember/recall pairs, while 3d dist2 would require 3). 

I hope a future implementation allows a more holistic approach over all assets in an achievement set, similar to how RAScript via RATools currently works, and really that the onus for poorly implemented iteration constructs (and bad code in general) is put onto CRs and D/C, especially because any hangs could theoretically just be interrupted by rcheevos itself as it would have control over the lua state.

I think something which should be considered is that lua could actually offer performance benefits for more complicated titles since every possible runtime case has to be solved by the rcheevos logic while logic with branches could just take the correct path first. Additionally, work cannot be shared across assets currently, meaning any calculations which could have been shared have to be done for every single asset, and for every single frame. A good example of this pitfall befalls one of my sets, Sega Rally Championship 2 which, since it runs on the Windows CE framework, every achievement has to have many superfluous alt groups where they just heuristically guess where the game process starts in memory, and where dynamic variables are located within the memory. These checks have to occur over and over, for every asset, every single frame, while with something like lua this could be calculated once and then used ad infinitum for the same session.

Jamiras:
> a very popular example would be for example calculating 2d dist2,

This can be done without using any intermediary variables:

```
sqrt((x2-x1)² + (y2-y1)²) < n
(x2-x1)² + (y2-y1)² < n²
(x2-x1)(x2-x1) + (y2-y1)(y2-y1) < n²
(x2*x2)+(x2*-x1)+(-x1*x2)+(-x1*-x1) + (y2*y2)+(y2*-y1)+(-y1*y2)+(-y1*-y1) < n²
```
```
AddSource x2 * x2
SubSource x2 * x1
SubSource x1 * x2
AddSource x1 * x1
AddSource y2 * y2
SubSource y2 * y1
SubSource y1 * y2
AddSource y1* y1
          0 < 100  # or whatever n*n is
```
I admit it's not clear that that is doing a 2D distance check, but anything with more than a couple of chained conditions can be hard to grok.

I haven't tried the 3D distance calculation, but I imagine it has a similar expansion.

Jamiras:
> Additionally, work cannot be shared across assets currently

I am totally for doing this, it just requires a lot of work across all levels of the architecture and there so many other higher priority and/or easier things to knock out first.

I'd like to define variables as a new asset type. They would be built like a leaderboard value. The result could be used in any other asset, including additional variables. They would also have their own delta/prior value, so you could do something like:

```
* {pokemon_caught}: (Measured large AddSource chain)
* achievement1 "catch em all": {pokemon_caught}=251, delta {pokemon_caught}=251
* achievement2 "catch 100 of em": {pokemon_caught}=100, delta {pokemon_caught}=100
* rich presence: "@Number({pokemon_caught}) pokemon caught"
```

Souzooka:
I hadn't considered looking for such an expansion for 2d dist2, I'll have to try throwing that into logic sometime. One limitation though is that it wouldn't work if both x1 and x2 were behind two different pointers (i.e. you were doing a 2d distance check between two different game entities) though I imagine x2,y2 would usually be a hardcoded value most of the time anyways.

PR #416, opened 2025-04-07
WiiWare Hash

LillyJadeKatrin:
See https://wiibrew.org/wiki/WAD_files for documentation on WAD file structure. Advice from Dolphin team was to hash the TMD and data portions of the WAD.

### DIFF
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -1117,15 +1117,14 @@ static int rc_hash_gamecube(char hash[33], const rc_hash_iterator_t* iterator)
   return success == 0 ? 0 : rc_hash_finalize(iterator, &md5, hash);
 }
 
-static int rc_hash_wii(char hash[33], const rc_hash_iterator_t* iterator)
+static int rc_hash_wii_disc(md5_state_t* md5,
+                            const rc_hash_iterator_t* iterator,
+                            void* file_handle)
 {
   const uint32_t MAIN_HEADER_SIZE = 0x80;
   const uint64_t REGION_CODE_ADDRESS = 0x4E000;
   const uint32_t CLUSTER_SIZE = 0x7C00;
   const uint32_t MAX_CLUSTER_COUNT = 1024;
-
-  md5_state_t md5;
-  void* file_handle;
 
   uint32_t partition_info_table[8];
   uint8_t total_partition_count = 0;
@@ -1141,20 +1140,6 @@ static int rc_hash_wii(char hash[33], const rc_hash_iterator_t* iterator)
 
   uint32_t ix, jx, kx;
   uint8_t encrypted;
-
-  file_handle = rc_file_open(iterator, iterator->path);
-  if (!file_handle)
-    return rc_hash_iterator_error(iterator, "Could not open file");
-
-  md5_init(&md5);
-  /* Check Magic Word */
-  rc_file_seek(iterator, file_handle, 0x18, SEEK_SET);
-  rc_file_read(iterator, file_handle, quad_buffer, 4);
-  if (!(quad_buffer[0] == 0x5D && quad_buffer[1] == 0x1C && quad_buffer[2] == 0x9E && quad_buffer[3] == 0xA3))
-  {
-    rc_file_close(iterator, file_handle);
-    return rc_hash_iterator_error(iterator, "Not a Wii disc");
-  }
 
   /* Check encryption byte - if 0x61 is 0, disc is encrypted */
   rc_file_seek(iterator, file_handle, 0x61, SEEK_SET);
@@ -1173,12 +1158,12 @@ static int rc_hash_wii(char hash[33], const rc_hash_iterator_t* iterator)
 
   rc_hash_iterator_verbose_formatted(iterator, "Hashing %u byte main header for [%c%c%c%c%c%c]",
                                      MAIN_HEADER_SIZE, buffer[0], buffer[1], buffer[2], buffer[3], buffer[4], buffer[5]);
-  md5_append(&md5, buffer, MAIN_HEADER_SIZE);
+  md5_append(md5, buffer, MAIN_HEADER_SIZE);
 
   /* Hash region code */
   rc_file_seek(iterator, file_handle, REGION_CODE_ADDRESS, SEEK_SET);
   rc_file_read(iterator, file_handle, quad_buffer, 4);
-  md5_append(&md5, quad_buffer, 4);
+  md5_append(md5, quad_buffer, 4);
 
   /* Scan partition table */
   rc_file_seek(iterator, file_handle, 0x40000, SEEK_SET);
@@ -1230,7 +1215,7 @@ static int rc_hash_wii(char hash[33], const rc_hash_iterator_t* iterator)
     rc_file_read(iterator, file_handle, buffer, tmd_size);
     rc_hash_iterator_verbose_formatted(iterator, "Hashing %u byte title metadata (partition type %u)",
                                        tmd_size, partition_table[jx + 1]);
-    md5_append(&md5, buffer, tmd_size);
+    md5_append(md5, buffer, tmd_size);
 
     /* Hash partition */
     rc_file_seek(iterator, file_handle, ((uint64_t)partition_table[jx] << 2) + 0x2B8, SEEK_SET);
@@ -1251,12 +1236,12 @@ static int rc_hash_wii(char hash[33], const rc_hash_iterator_t* iterator)
       {
         rc_file_seek(iterator, file_handle, part_offset + (ix * 0x8000) + 0x400, SEEK_SET);
         rc_file_read(iterator, file_handle, buffer, CLUSTER_SIZE);
-        md5_append(&md5, buffer, CLUSTER_SIZE);
+        md5_append(md5, buffer, CLUSTER_SIZE);
       }
     }
     else /* Decrypted */
     {
-      if (rc_hash_nintendo_disc_partition(&md5, iterator, file_handle, part_offset, 2) == 0)
+      if (rc_hash_nintendo_disc_partition(md5, iterator, file_handle, part_offset, 2) == 0)
       {
         free(partition_table);
         free(buffer);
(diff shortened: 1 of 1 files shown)


src/rhash/hash.c
```diff
@@ -1223,6 +1223,51 @@ static int rc_hash_wii(md5_state_t* md5,
...
+  uint8_t quad_buffer[4];
+  uint8_t* buffer;
+
+  /* GetApploaderSize */
+  rc_file_seek(iterator, file_handle, 0x08, SEEK_SET);
+  rc_file_read(iterator, file_handle, quad_buffer, 4);
+  cert_chain_size =
+    ((uint32_t)quad_buffer[0] << 24) | (quad_buffer[1] << 16) | (quad_buffer[2] << 8) | quad_buffer[3];
+  /* Round up to nearest 0x40 */
+  cert_chain_size = ((cert_chain_size - 1) / 0x40 + 1) * 0x40;
```
Jamiras, line 1242: This can be done more efficiently:
```
cert_chain_size = (cert_chain_size + 0x3F) & ~0x3F;
```

src/rhash/hash.c
```diff
@@ -1223,6 +1223,51 @@ static int rc_hash_wii(md5_state_t* md5,
...
+  tmd_size =
+    ((uint32_t)quad_buffer[0] << 24) | (quad_buffer[1] << 16) | (quad_buffer[2] << 8) | quad_buffer[3];
+  tmd_size = ((tmd_size - 1) / 0x40 + 1) * 0x40;
+  rc_file_read(iterator, file_handle, quad_buffer, 4);
+  data_size =
+    ((uint32_t)quad_buffer[0] << 24) | (quad_buffer[1] << 16) | (quad_buffer[2] << 8) | quad_buffer[3];
+  data_size = ((data_size - 1) / 0x40 + 1) * 0x40;
+
+  tmd_start_addr = 0x20 + cert_chain_size + ticket_size;
+  hash_size = tmd_size + data_size;
```
Jamiras, line 1258: recommend having an upper limit here in case something that looks like a WiiWare file gets through and the sizes are garbage.
```
if (hash_size > MAX_BUFFER_SIZE)
    hash_size = MAX_BUFFER_SIZE;
```

src/rhash/hash.c
```diff
@@ -1250,7 +1295,16 @@ static int rc_hash_nintendo_disc(char hash[33], const rc_hash_iterator_t* iterat
...
+      /* Check WiiWare */
+      rc_file_seek(iterator, file_handle, 0x04, SEEK_SET);
+      rc_file_read(iterator, file_handle, quad_buffer, 2);
+      if (quad_buffer[0] == 'I' && quad_buffer[1] == 's')
+      {
+        success = rc_hash_wiiware(&md5, iterator, file_handle);
+      }
+      else {
+        success = rc_hash_iterator_error(iterator, "Not a supported Nintendo disc");
+      }
```
Jamiras, line 1307: make sure to add an entry for `wad` files in the `rc_hash_iterator_ext_handlers` table.

src/rhash/hash.c
```diff
@@ -1250,7 +1295,16 @@ static int rc_hash_nintendo_disc(char hash[33], const rc_hash_iterator_t* iterat
     {
       success = rc_hash_wii(&md5, iterator, file_handle);
     } else {
-      success = rc_hash_iterator_error(iterator, "Not a supported Nintendo disc");
+      /* Check WiiWare */
+      rc_file_seek(iterator, file_handle, 0x04, SEEK_SET);
+      rc_file_read(iterator, file_handle, quad_buffer, 2);
+      if (quad_buffer[0] == 'I' && quad_buffer[1] == 's')
```
Jamiras, line 1301: Consider also validating buffer[2] and [3], which I believe both should be 0x00 per the linked document. This will reduce the likelihood of false positives.

src/rhash/hash.c
```diff
@@ -1223,6 +1223,51 @@ static int rc_hash_wii(md5_state_t* md5,
...
+    ((uint32_t)quad_buffer[0] << 24) | (quad_buffer[1] << 16) | (quad_buffer[2] << 8) | quad_buffer[3];
+  ticket_size = ((ticket_size - 1) / 0x40 + 1) * 0x40;
+  rc_file_read(iterator, file_handle, quad_buffer, 4);
+  tmd_size =
+    ((uint32_t)quad_buffer[0] << 24) | (quad_buffer[1] << 16) | (quad_buffer[2] << 8) | quad_buffer[3];
+  tmd_size = ((tmd_size - 1) / 0x40 + 1) * 0x40;
+  rc_file_read(iterator, file_handle, quad_buffer, 4);
+  data_size =
+    ((uint32_t)quad_buffer[0] << 24) | (quad_buffer[1] << 16) | (quad_buffer[2] << 8) | quad_buffer[3];
+  data_size = ((data_size - 1) / 0x40 + 1) * 0x40;
```
Jamiras, line 1255: I understand that sections are aligned to 0x40 addresses, but is the padding between the actual data_size and the next 0x40 boundary guaranteed to be the same in all dumps? Should we just hash `tmd_size` bytes, then advance to the next boundary and hash `data_size` bytes?

Jamiras (commented):
An important thing to check for any new hash algorithm is consistency. Since you didn't provide any hashes that you generated, here's some that I generated:
| title | hash |
| ---- | ---- |
| Adventure Island - The Beginning (USA) | 5e22f89a0a3d8a3ae03da23aac4fa93a |
| Cave Story (USA) (v258) | 96b8225d690f6b2d93ec48a53d4abac6 |
| Mega Man 9 (USA) | 4495a201b6e9e6ac964e7e10fde5b0b2 |
| Tales of Monkey Island - Chapter 1 | 8ae653c9efb761e69765c06b16a3e5c9 |

LillyJadeKatrin:
> An important thing to check for any new hash algorithm is consistency. Since you didn't provide any hashes that you generated, here's some that I generated:
> 
> title	hash
> Adventure Island - The Beginning (USA)	5e22f89a0a3d8a3ae03da23aac4fa93a
> Cave Story (USA) (v258)	96b8225d690f6b2d93ec48a53d4abac6
> Mega Man 9 (USA)	4495a201b6e9e6ac964e7e10fde5b0b2
> Tales of Monkey Island - Chapter 1	8ae653c9efb761e69765c06b16a3e5c9

I actually did but none of these - they're on a second tab in the excel file I used for the Wii hashes. https://docs.google.com/spreadsheets/d/1tiqr2qXQ6Ovshq28Pqr97j98ulgwFpr_Dt6WFf3E3PM/edit?usp=drivesdk

src/rhash/hash.c
LillyJadeKatrin: Question has been posed to a Dolphin developer with more experience.

src/rhash/hash.c
```diff
@@ -1223,6 +1223,53 @@ static int rc_hash_wii(md5_state_t* md5,
...
+                           const rc_hash_iterator_t* iterator,
+                           void* file_handle)
+{
+  uint32_t cert_chain_size, ticket_size, tmd_size, data_size;
+  uint32_t tmd_start_addr, hash_size;
+
+  uint8_t quad_buffer[4];
+  uint8_t* buffer;
+
+  /* GetApploaderSize */
```
JosJuice, line 1236: This is reading the certificate chain size, not the apploader size.

src/rhash/hash.c
```diff
@@ -1223,6 +1223,53 @@ static int rc_hash_wii(md5_state_t* md5,
...
+  rc_file_seek(iterator, file_handle, 0x08, SEEK_SET);
+  rc_file_read(iterator, file_handle, quad_buffer, 4);
+  cert_chain_size =
+    ((uint32_t)quad_buffer[0] << 24) | (quad_buffer[1] << 16) | (quad_buffer[2] << 8) | quad_buffer[3];
```
JosJuice, line 1240: The cast to `uint32_t` is redundant (if you don't support `sizeof(int) < 4`), or you need to cast to `uint32_t` before doing `<< 16` (if you do support `sizeof(int) < 4`). Same below.

src/rhash/hash.c
JosJuice: I'm not sure if the padding is always the same. I would recommend skipping the padding to be on the safe side, but this will make things a bit more complicated for you, because:
* For the content data, you should skip not only the padding after the last content, but the padding between each content.
* Because the content data is encrypted, the size to hash for each content should be rounded up to the size of an AES block.

src/rhash/hash.c
```diff
@@ -1250,7 +1297,16 @@ static int rc_hash_nintendo_disc(char hash[33], const rc_hash_iterator_t* iterat
...
     } else {
-      success = rc_hash_iterator_error(iterator, "Not a supported Nintendo disc");
+      /* Check WiiWare */
+      rc_file_seek(iterator, file_handle, 0x04, SEEK_SET);
+      rc_file_read(iterator, file_handle, quad_buffer, 4);
+      if (quad_buffer[0] == 'I' && quad_buffer[1] == 's' && quad_buffer[2] == 0x00 && quad_buffer[3] == 0x00)
+      {
+        success = rc_hash_wiiware(&md5, iterator, file_handle);
+      }
+      else {
```
JosJuice, line 1307: Inconsistent brace style.

src/rhash/hash.c
```diff
@@ -1223,41 +1270,93 @@ static int rc_hash_wii(md5_state_t* md5,
...
-    } else {
-      success = rc_hash_iterator_error(iterator, "Not a supported Nintendo disc");
+      content_size <<= 8;
+      content_size += oct_buffer[jx];
     }
+    content_size = (content_size + 0x0F) & ~0x0F;
+    if (content_size > MAX_BUFFER_SIZE)
+      content_size = MAX_BUFFER_SIZE;
+    /* Hash content */
+    buffer = (uint8_t*)malloc(tmd_size);
```
JosJuice, line 1349:
```suggestion
    buffer = (uint8_t*)malloc(content_size);
```

include/rc_consoles.h
```diff
@@ -94,6 +94,7 @@ enum {
   RC_CONSOLE_TI83 = 79,
   RC_CONSOLE_UZEBOX = 80,
   RC_CONSOLE_FAMICOM_DISK_SYSTEM = 81,
+  RC_CONSOLE_WIIWARE = 82,
```
Jamiras, line 97: I don't think this deserves a separate console id. WiiWare is just a different way to distribute games for Wii, just like [DSIWare](https://retroachievements.org/hub/20698) is for DSi.

Or disk vs cartridge for something like the Commodore 64 or MSX.

src/rhash/hash.c
```diff
@@ -1205,59 +1251,113 @@ static int rc_hash_wii(md5_state_t* md5,
...
-    rc_file_read(iterator, file_handle, quad_buffer, 4);
-    if (quad_buffer[0] == 0x5D && quad_buffer[1] == 0x1C && quad_buffer[2] == 0x9E && quad_buffer[3] == 0xA3)
+    rc_file_close(iterator, file_handle);
+    return rc_hash_iterator_error(iterator, "Not a supported WiiWare wad");
+  }
+
+  rc_file_seek(iterator, file_handle, 0x08, SEEK_SET);
+  rc_file_read(iterator, file_handle, quad_buffer, 4);
+  cert_chain_size =
+    ((uint32_t)quad_buffer[0] << 24) | ((uint32_t)quad_buffer[1] << 16) | ((uint32_t)quad_buffer[2] << 8) | (uint32_t)quad_buffer[3];
```
Jamiras, line 1304: I thought we were going to remove the explicit casts, not add more of them.

src/rhash/hash.c
```diff
@@ -1205,59 +1251,113 @@ static int rc_hash_wii(md5_state_t* md5,
...
+    content_size = 0;
+    for (jx = 0; jx < 8; jx++)
     {
-      success = rc_hash_wii(&md5, iterator, file_handle);
-    } else {
-      success = rc_hash_iterator_error(iterator, "Not a supported Nintendo disc");
+      content_size <<= 8;
+      content_size += oct_buffer[jx];
     }
+    content_size = (content_size + 0x0F) & ~0x0F;
```
Jamiras, line 1346: Force pushing makes it very hard to look at changes since I last reviewed.

I thought we were going to get rid of the padding adjustment, not change it from 0x3F to 0x0F.

src/rhash/hash.c
```diff
@@ -1205,59 +1251,113 @@ static int rc_hash_wii(md5_state_t* md5,
...
+    rc_file_seek(iterator, file_handle, (uint64_t)tmd_start_addr + 0x1e4 + 8 + ix * 0x20, SEEK_SET);
+    rc_file_read(iterator, file_handle, oct_buffer, 8);
+    content_size = 0;
+    for (jx = 0; jx < 8; jx++)
     {
-      success = rc_hash_wii(&md5, iterator, file_handle);
-    } else {
-      success = rc_hash_iterator_error(iterator, "Not a supported Nintendo disc");
+      content_size <<= 8;
+      content_size += oct_buffer[jx];
```
Jamiras, line 1344: `content_size` is a `uint32_t`, and won't hold four bytes of shifted data.

Also, why build it with a loop instead of the shifting+or logic used for `tmd_size` (and others)?

Since we're limiting the buffer size to 64MB, you really just need to check if the upper 4 bytes are non-zero and force the maximum size. None of my test files event hit the 64MB limit, are there really wad files bigger than 4GB?

src/rhash/hash.c
```diff
@@ -1205,59 +1251,113 @@ static int rc_hash_wii(md5_state_t* md5,
...
+    content_size = (content_size + 0x0F) & ~0x0F;
+    if (content_size > MAX_BUFFER_SIZE)
+      content_size = MAX_BUFFER_SIZE;
+    /* Hash content */
+    buffer = (uint8_t*)malloc(content_size);
+    rc_file_seek(iterator, file_handle, content_addr, SEEK_SET);
+    rc_file_read(iterator, file_handle, buffer, content_size);
+    md5_append(&md5, buffer, content_size);
+    content_addr += content_size;
+    content_addr = (content_addr + 0x3F) & ~0x3F;
```
Jamiras, line 1355: Minor detail: You're applying a ceiling to the address, not the content size. If the address was not already aligned to an 0x40 boundary, this will introduce an unintended drift.

src/rhash/hash.c
LillyJadeKatrin: This one in particular is aligned to the nearest 0x10 on direction from Jos pertaining to the individual content files, my understanding of what they told me was that each file aligns to 0x10 and the entire section aligns to 0x40 in total.

JosJuice: Each content is aligned to 0x40, but I suggested that the padding shouldn't be hashed in its entirety up to the 0x40 mark, only to the 0x10 mark (the size of an AES block).

src/rhash/hash.c
JosJuice: All addresses in WAD files are aligned to 0x40. If `content_addr` were to somehow become not aligned to 0x40 (which shouldn't be possible the way the code is written), then keeping it misaligned seems like the worse option to me.

src/rhash/hash.c
LillyJadeKatrin: > Also, why build it with a loop instead of the shifting+or logic used for `tmd_size` (and others)?
Because I didn't want to do that for an eight-byte value, honestly. I did a search for wad file sizes and I found that both Dragon Quest X and Rock Band had DLC wads exceeding 4GB but neither of those would be hashed this way even without the 64MB limit.

src/rhash/hash.c
Jamiras: This probably deserves a comment:
```
/* Padding between content should be ignored. But because the content data is encrypted, the size to hash for each content should be rounded up to the size of an AES block (16 bytes). */
```
Similarly, the `/* Round up to nearest 0x40 */` comment should be expanded:
```
/* Each content is individually aligned to a 0x40-byte boundary. */
```

src/rhash/hash.c
```diff
@@ -1223,7 +1255,83 @@ static int rc_hash_wii(md5_state_t* md5,
...
+    else
+    {
+      content_size = MAX_BUFFER_SIZE;
+    }
+    /* Hash content */
+    buffer = (uint8_t*)malloc(content_size);
+    rc_file_seek(iterator, file_handle, content_addr, SEEK_SET);
+    rc_file_read(iterator, file_handle, buffer, content_size);
+    md5_append(md5, buffer, content_size);
+    content_addr += content_size;
```
Jamiras, line 1326: You'll want to store the original `content_size` somewhere. If it happens to be greater than `MAX_BUFFER_SIZE`, you'll only advance by `MAX_BUFFER_SIZE`.

Jamiras:
None of my hashes match your document. Has it been updated since we changed the padding logic?
```
afd0763c3fd3cd43b13fe88bda01b12f        Adventure Island - The Beginning (USA) (WiiWare).wad
05b1168a16ca1c5980ecb2bc82b065a1        Cave Story (USA) (v258) (WiiWare).wad
08aaef52405f06651fc3150e4ce8413b        Mega Man 9 (USA) (WiiWare).wad
b8d0b09059e6c12476fe7ade779c8a31        Phoenix Wright - Ace Attorney (USA) (WiiWare).wad
7bede236b53452a8718a057ae849f4ee        Sonic the Hedgehog 4 - Episode I (USA) (WiiWare).wad
2851f77e1b967bd5a96849870452479f        Tales of Monkey Island - Chapter 1 - Launch of the Screaming Narwhal (USA) (WiiWare).wad
92d6d1b11cfdeb718b634205b3f2c061        UNO (USA) (WiiWare).wad
```

LillyJadeKatrin:
Huh. No, I hadn't, but I was evidently generating them differently from you.

I have a local branch of Dolphin that, in theory, is set up to be rebased on top of a submodule update so that WiiWare games can be supported, and I was getting my hashes from the resulting Dolphin logs. However, when I take the time to update RAHasher properly with this algorithm, I get different hashes.

I've updated the doc with the hashes I get from RAHasher instead, and the three WADs we have in common do match this way. I'll investigate what I might have done with Dolphin that's causing it to hash differently than RAHasher, but that's a different repo entirely.

LillyJadeKatrin:
Found some odd behavior in the hash algorithm that seems to be making it think files are much bigger than they are. Drafting this until I nail it down.

LillyJadeKatrin:
Found an error in the algorithm, fixed it, now I get the same hashes when I include this in RAHasher and in Dolphin, the hash document has been updated with the new hashes.

Jamiras:
My hashes:
```
89ea4766f5d79f67981d225595e0ceca        Adventure Island - The Beginning (USA) (WiiWare).wad
e7ad0e731d0c933a9ee814fc34a8b48a        Cave Story (USA) (v258) (WiiWare).wad
088a541ccdc2d906d6b432cfca7a7959        Mega Man 9 (USA) (WiiWare).wad
28938c0b385c0db9c559fe4affdb9e6a        Phoenix Wright - Ace Attorney (USA) (WiiWare).wad
e0a92032201ad646713d5d7f4abc2099        Sonic the Hedgehog 4 - Episode I (USA) (WiiWare).wad
a169143caf19fd6c2c24b06b4ba0e0b1        Tales of Monkey Island - Chapter 1 - Launch of the Screaming Narwhal (USA) (WiiWare).wad
ddbc71da3f48c51c1f3450d68f8b0440        UNO (USA) (WiiWare).wad
```
The three that are in your document match mine.

Jamiras (approved):

PR #432, opened 2025-05-11
Add allprogress/hashlibrary endpoint wrappers

stenzek:
These functions when paired together allow an emulator to populate its game list with the achievement count for each game, and the number of achievements unlocked by the user.

NOTE: Consumers of these APIs should cache the responses to these calls on disk. You should *not* issue an allprogress call every time the application starts or the list refreshes.

### DIFF
--- a/include/rc_client.h
+++ b/include/rc_client.h
@@ -219,6 +219,39 @@ typedef struct rc_client_user_game_summary_t {
  */
 RC_EXPORT void RC_CCONV rc_client_get_user_game_summary(const rc_client_t* client, rc_client_user_game_summary_t* summary);
 
+typedef struct rc_client_all_user_progress_entry_t {
+  uint32_t game_id;
+  uint32_t num_achievements;
+  uint32_t num_unlocked_achievements;
+  uint32_t num_unlocked_achievements_hardcore;
+} rc_client_all_user_progress_entry_t;
+
+typedef struct rc_client_all_user_progress_t {
+  rc_client_all_user_progress_entry_t* entries;
+  uint32_t num_entries;
+} rc_client_all_user_progress_t;
+
+/**
+ * Callback that is fired when an all progress query completes. list may be null if the query failed.
+ */
+typedef void(RC_CCONV* rc_client_fetch_all_user_progress_callback_t)(int result, const char* error_message,
+                                                                     rc_client_all_user_progress_t* list,
+                                                                     rc_client_t* client, void* callback_userdata);
+
+/**
+ * Starts an asynchronous request for all progress for the given console.
+ * This query returns the total number of achievements for all games tracked by this console, as well as
+ * the user's achievement unlock count for both softcore and hardcore modes.
+ */
+RC_EXPORT rc_client_async_handle_t* RC_CCONV
+rc_client_begin_fetch_all_user_progress(rc_client_t* client, uint32_t console_id,
+                                        rc_client_fetch_all_user_progress_callback_t callback, void* callback_userdata);
+
+/**
+ * Destroys a previously-allocated result from the rc_client_begin_fetch_all_progress_list() callback.
+ */
+RC_EXPORT void RC_CCONV rc_client_destroy_all_user_progress(rc_client_all_user_progress_t* list);
+
 /*****************************************************************************\
 | Game                                                                        |
 \*****************************************************************************/
@@ -316,6 +349,40 @@ typedef struct rc_client_subset_t {
 } rc_client_subset_t;
 
 RC_EXPORT const rc_client_subset_t* RC_CCONV rc_client_get_subset_info(rc_client_t* client, uint32_t subset_id);
+
+/*****************************************************************************\
+| Fetch Game Hashes                                                           |
+\*****************************************************************************/
+
+typedef struct rc_client_hash_library_entry_t {
+  char hash[33];
+  uint32_t game_id;
+} rc_client_hash_library_entry_t;
+
+typedef struct rc_client_hash_library_t {
+  rc_client_hash_library_entry_t* entries;
+  uint32_t num_entries;
+} rc_client_hash_library_t;
+
+/**
+ * Callback that is fired when a hash library request completes. list may be null if the query failed.
+ */
+typedef void(RC_CCONV* rc_client_fetch_hash_library_callback_t)(int result, const char* error_message,
+                                                                rc_client_hash_library_t* list, rc_client_t* client,
+                                                                void* callback_userdata);
+
+/**
+ * Starts an asynchronous request for all hashes for the given console.
+ * This request returns a mapping from hashes to the game's unique identifier. A single game may have multiple
+ * hashes in the case of multi-disc games, or variants that are still compatible with the same achievement set.
+ */
+RC_EXPORT rc_client_async_handle_t* RC_CCONV rc_client_begin_fetch_hash_library(
+  rc_client_t* client, uint32_t console_id, rc_client_fetch_hash_library_callback_t callback, void* callback_userdata);
+
+/**
+ * Destroys a previously-allocated result from the rc_client_destroy_hash_library() callback.
+ */
+RC_EXPORT void RC_CCONV rc_client_destroy_hash_library(rc_client_hash_library_t* list);
 
 /*****************************************************************************\
 | Achievements                                                                |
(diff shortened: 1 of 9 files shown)


include/rc_client.h
```diff
@@ -317,6 +317,91 @@ typedef struct rc_client_subset_t {
...
+  rc_client_t* client, uint32_t console_id, rc_client_fetch_hash_library_callback_t callback, void* callback_userdata);
+
+/**
+ * Destroys a previously-allocated result from the rc_client_destroy_hash_library() callback.
+ */
+RC_EXPORT void RC_CCONV rc_client_destroy_hash_library(rc_client_hash_library_t* list);
+
+/*****************************************************************************\
+| Game Info                                                                   |
+\*****************************************************************************/
```
Jamiras, line 362: This should probably just go into the "User" section (around line 220)

include/rc_client.h
```diff
@@ -317,6 +317,91 @@ typedef struct rc_client_subset_t {
 
 RC_EXPORT const rc_client_subset_t* RC_CCONV rc_client_get_subset_info(rc_client_t* client, uint32_t subset_id);
 
+/*****************************************************************************\
+| Fetch Game Hashes                                                           |
+\*****************************************************************************/
+
+typedef struct rc_client_hash_library_entry_t
+{
```
Jamiras, line 325: Opening brace should be on same line as type name (applies to other structs in this file)

include/rc_client.h
```diff
@@ -317,6 +317,91 @@ typedef struct rc_client_subset_t {
...
+  /* The hash for the game */
```
Jamiras, line 326: Per field comments are not provided in this file. The names should be descriptive enough to not require a comment (if additional detail is needed, it can be included in the wiki). And it allows for delineating version differences.
```
  uint8_t state;
  uint8_t category;
  uint8_t bucket;
  uint8_t unlocked;
  /* minimum version: 11.1 */
  float rarity;
  float rarity_hardcore;
  uint8_t type;
  /* minimum version: 12.0 */
  const char* badge_url;
  const char* badge_locked_url;
```

src/rapi/rc_api_info.c
```diff
@@ -492,3 +492,91 @@ int rc_api_process_fetch_game_titles_server_response(rc_api_fetch_game_titles_re
...
+                                                  const rc_api_fetch_hash_library_request_t* api_params,
+                                                  const rc_api_host_t* host)
+{
+  rc_api_url_builder_t builder;
+  rc_api_url_build_dorequest_url(request, host);
+
+  /* note: unauthenticated request */
+  rc_url_builder_init(&builder, &request->buffer, 48);
+  rc_url_builder_append_str_param(&builder, "r", "hashlibrary");
+  rc_url_builder_append_unum_param(&builder, "c", api_params->console_id);
```
Jamiras, line 514: It's more correct to not send the `c` parameter instead of sending 0:
```
if (api_params->console_id != RC_CONSOLE_UNKNOWN)
  rc_url_builder_append_unum_param(&builder, "c", api_params->console_id);
```

src/rapi/rc_api_info.c
```diff
@@ -492,3 +492,91 @@ int rc_api_process_fetch_game_titles_server_response(rc_api_fetch_game_titles_re
...
+
+  if (!fields[2].value_start) {
+    /* call rc_json_get_required_object to generate the error message */
+    rc_json_get_required_object(NULL, 0, &response->response, &fields[2], "MD5List");
+    return RC_MISSING_VALUE;
+  }
+
+  response->num_entries = fields[2].array_size;
+  if (response->num_entries > 0) {
+    rc_buffer_reserve(&response->response.buffer, response->num_entries * (32 + sizeof(rc_api_hash_library_entry_t)));
```
Jamiras, line 552: This should be `33+` (for the null terminator), but I wonder if it wouldn't make more sense to change the pointer into a buffer.
```
typedef struct rc_api_hash_library_entry_t {
  /* The hash for the game */
  char hash[33];
  /* The unique identifier of the game */
  uint32_t game_id;
} rc_api_hash_library_entry_t;
```

src/rapi/rc_api_info.c
```diff
@@ -492,3 +492,91 @@ int rc_api_process_fetch_game_titles_server_response(rc_api_fetch_game_titles_re
...
+    if (!response->entries)
+      return RC_OUT_OF_MEMORY;
+
+    memset(&iterator, 0, sizeof(iterator));
+    iterator.json = fields[2].value_start;
+    iterator.end = fields[2].value_end;
+
+    entry = response->entries;
+    while (rc_json_get_next_object_field(&iterator, &field)) {
+      /* TODO: This isn't handling escape characters in the key, the RC JSON parsing functions have no method for it. */
```
Jamiras, line 565: The keys should be hashes. What escape characters need to be handled?

src/rc_client.c
```diff
@@ -3247,6 +3247,275 @@ const rc_client_subset_t* rc_client_get_subset_info(rc_client_t* client, uint32_
...
+  rc_client_fetch_hash_library_callback_data_t* hashlib_callback_data =
+    (rc_client_fetch_hash_library_callback_data_t*)callback_data;
+  rc_client_t* client = hashlib_callback_data->client;
+  rc_api_fetch_hash_library_response_t hashlib_response;
+  const char* error_message;
+  int result;
+
+  result = rc_client_end_async(client, &hashlib_callback_data->async_handle);
+  if (result)
+  {
```
Jamiras, line 3272: Opening braces should not be on a separate line (except for the opening brace of the function - which isn't consistent between files). 

I wonder if there's something wrong with the `.editorconfig`, which does say `curly_bracket_next_line  = false`, but I often have to manually correct my files too. Maybe adding an `indent_brace_style` would help. I'll play around with this separately.

test/rapi/test_rc_api_info.c
```diff
@@ -425,6 +425,77 @@ static void test_process_fetch_game_titles_response() {
...
+  ASSERT_NUM_EQUALS(entry->game_id, 2);
+  ASSERT_STR_EQUALS(entry->hash, "8899aabbccddeeff0011223344556677");
+
+  rc_api_destroy_fetch_hash_library_response(&fetch_hash_library_response);
+}
+
+static void test_process_fetch_hash_library_empty_response() {
+  rc_api_fetch_hash_library_response_t fetch_hash_library_response;
+  rc_api_server_response_t fetch_hash_library_server_response;
+  const char* server_response = "{\"Success\":true,\"MD5List\":[]}"; /* RAWeb returns a list instead of a map for invalid console */
```
Jamiras, line 481: I plan to correct that, but it's good to ensure the current behavior is functional.

src/rc_client.c
stenzek: It's probably because I was editing it within the DS tree, and it was picking up my `clang-format` in the root there. So likely no issue on your end.

src/rapi/rc_api_info.c
stenzek: I changed the `rc_client` side, but there seems to be a few other places (e.g. badge names) where a variable length is used in rapi, and a fixed-size buffer in `rc_client`. Any preference?

src/rapi/rc_api_info.c
stenzek: Can't remember for certain since I wrote this months ago, I think it was just a note because the only other places that iterate a map immediately convert the key to an integer, so there will never be any escaped characters. Whereas the key is treated as a string here, and theoretically there could be.

Practically speaking, since you control the server, probably not. Happy to drop the comment if you like.

src/rapi/rc_api_info.c
Jamiras: My preference is fixed-sized buffers for strings of fixed length and shorter strings (to eliminate the overhead of the pointer) - particularly for things that will stay in memory for an extended period of time.

`rapi` uses pointers for everything to simplify the JSON parsing interactions and to better support unexpected data (null/missing values or strings longer than expected). `rapi` also allocates chunks of memory to avoid smaller allocations, so it'll generally use more memory than it needs, but the responses should be relatively short-lived, so the overhead is less important there.

Jamiras: The comment about using 33 in your estimate still applies.

src/rapi/rc_api_info.c
Jamiras: I'm generally not fond of using generic key maps in JSON. If this wasn't a long-standing pre-existing API, I'd have just made it an array of pairs (like the output) instead of a map.

And as long as keys aren't coming from user data, they should never need escaping. While the comment may be applicable in general, it's not applicable to the place where it's at. Please remove it.

Jamiras (approved):

PR #446, opened 2025-06-19
Add 3DS memory map

CasualPokePlayer:
The "real" addresses set here are based on 3DS virtual memory addresses, with this memory map intended to map 1:1 with 3DS userland virtual memory

Unsure if this would end up being desirable in practice. This has the benefit of making pointers map 1:1 to RetroAchievements memory space (which is a huge win given how pointer heavy 3DS games will be and the disjointed mess that "real" memory lies in the memory map). This will, however, balloon the memory space to search in for developers. BizHawk (the current 3DS RetroAchievements capable emulator), as an implementation detail, also slightly suffers for the achievement runtime as it will allocate 1GiB here for a memory block LUT; this is an implementation detail, however, just due to being a multi-system emulator, a standalone emulator wouldn't have such, and even so 1GiB of memory is relatively free for any machine that can handle 3DS emulation anyways.

A different memory map could potentially be made which only covers the "real" memory spaces, but doing so will still have a large amount of memory mapped (definitely more than 256MiB, the size of the New 3DS's FCRAM), and will be much harder for the current achievements implementation to handle pointers (given the reported problems with even handling MEM1/MEM2 on the Wii).


redwizard42:
Does this mean that if I am searching RAM and find a pointer that points to the new linear heap , it'll have a value in the 0x30000000-0x37FFFFFF range?

CasualPokePlayer:
> Does this mean that if I am searching RAM and find a pointer that points to the new linear heap , it'll have a value in the 0x30000000-0x37FFFFFF range?

That is usually correct (it's technically more 0x30000000-0x3FFFFFFF, only 0x30000000-0x37FFFFFF if you're on the old 3DS)

test/rcheevos/test_consoleinfo.c
```diff
@@ -163,6 +163,7 @@ void test_consoleinfo(void) {
   TEST_PARAMS2(test_memory, RC_CONSOLE_NEO_GEO_CD, 0x010000);
   TEST_PARAMS2(test_memory, RC_CONSOLE_NINTENDO, 0x010000);
   TEST_PARAMS2(test_memory, RC_CONSOLE_NINTENDO_64, 0x800000);
+  TEST_PARAMS2(test_memory, RC_CONSOLE_NINTENDO_3DS, 0x40000000);
```
Jamiras, line 166: This will need to be updated to account for changes in #462.
```
  TEST_PARAMS3(test_memory, RC_CONSOLE_NINTENDO_3DS, 0x40000000, 0x2CE00000);
```

I'm a bit weary that there's still 750MB of searchable memory given the system only has 256MB of RAM.

The Code Binary segment, IPC Buffers segment, and Shared Memory segments are each 64MB. The Regular Heap and Linear Heap segments are each 128MB. There's also another ~48MB of miscellaneous sections that aren't the primary 256MB of RAM.

Would achievement developers be expected to find data in all of those regions? Anything that's not marked as UNUSED is currently searchable. Would it make sense to add new enum values for non-searchable system ram? 

Maybe they could be marked as VIRTUAL_RAM and then have that made as non-searchable. Currently VIRTUAL_RAM is only being used for echo/mirror RAM segments, and eliminating those from search results would probably actually benefit developers on those systems.

src/rcheevos/consoleinfo.c
```diff
@@ -759,6 +759,29 @@ static const rc_memory_region_t _rc_memory_regions_nintendo_dsi[] = {
...
+    { 0x1E800000U, 0x1EBFFFFFU, 0x1E800000U, RC_MEMORY_TYPE_SYSTEM_RAM, "New 3DS Memory" },
+    { 0x1EC00000U, 0x1EFFFFFFU, 0x1EC00000U, RC_MEMORY_TYPE_HARDWARE_CONTROLLER, "I/O Registers" },
+    { 0x1F000000U, 0x1F5FFFFFU, 0x1F000000U, RC_MEMORY_TYPE_VIDEO_RAM, "VRAM" },
+    { 0x1F600000U, 0x1FEFFFFFU, 0x1F600000U, RC_MEMORY_TYPE_UNUSED, "" },
+    { 0x1FF00000U, 0x1FF7FFFFU, 0x1FF00000U, RC_MEMORY_TYPE_HARDWARE_CONTROLLER, "DSP Memory" },
+    { 0x1FF80000U, 0x1FF81FFFU, 0x1FF80000U, RC_MEMORY_TYPE_READONLY, "Configuration Memory" },
+    { 0x1FF82000U, 0x1FFFFFFFU, 0x1FF82000U, RC_MEMORY_TYPE_SYSTEM_RAM, "Thread Local Storage" }, /* Most of this is actually unused in practice */
+    { 0x20000000U, 0x2FFFFFFFU, 0x20000000U, RC_MEMORY_TYPE_UNUSED, "" },
+    { 0x30000000U, 0x37FFFFFFU, 0x30000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "New Linear Heap" }, /* Newer games use this as the linear heap base instead */
+    { 0x38000000U, 0x3FFFFFFFU, 0x38000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "New Linear Heap (New 3DS Exclusive)" } /* New 3DS exclusive space for the newer linear heap */
```
Jamiras, line 781: This map makes sense given the referenced documentation, but only these last two segments are the physical memory reported in the system specs. Where does all the other SYSTEM_RAM come from?

Jamiras (commented):
Sorry it took me so long to get to this. My summer was super busy, and I've been playing catch-up ever since.

test/rcheevos/test_consoleinfo.c
CasualPokePlayer: The Code Binary segment is something I would expect would be needed to looked at, as that doesn't just contain the .text, but also contains .data and .bss which are likely going to contain variables (and especially base pointers for fun pointer hell).

IPC Buffers and Shared Memory are unlikely to contain anything actually useful, as that's for handling communications with other 3DS processes (generally speaking this just means talking to the OS for various services, or in the case of emulators, the emulator HLE'ing those services). This is why those are listed as RC_MEMORY_TYPE_HARDWARE_CONTROLLER as they effectively work as "modern" MMIO (for unprivileged userland anyways).

src/rcheevos/consoleinfo.c
CasualPokePlayer: I'm not sure what you mean here. To be clear, not all of the ranges will necessarily be addressed to actual memory (they would be mapped on the fly as allocations happen), not even neccessarily contiguous in their mapping, they're just the maximum ranges (assuming such memory is available in practice, in practice you'd OOM before you'd be able to do that). Most of the ranges here would possibly be addressed to some part of the 128/256MiB FCRAM (including things like IPC Buffers and Shared Memory, notwithstanding them marked as RC_MEMORY_TYPE_HARDWARE_CONTROLLER).

Jamiras: The hardware specs for 3DS say 128MB of RAM, and the hardware specs for New3DS say 256MB of RAM. I'm just confused where the data in all the other blocks marked as SYSTEM_RAM is being stored.

If it's just mapping data from the "New Linear Heap" segments, it could be marked as VIRTUAL_RAM instead of SYSTEM_RAM.

Ideally, we need some way to restrict the searchable memory space. Otherwise, when a dev captures memory to look for changes, they'll have to capture the full 750MB for the non-UNUSED memory blocks, and there could be redundant matches where memor is mirrored. Then, we would need some way to guide the developer into preferring one over an other.

CasualPokePlayer: Again, not all of the range is necessarily going to be mapped, and there's no guarantee different mappings will be contiguous. Linear Heap allocations won't be overlapping with other memory allocations, so ranges where other memory has already been allocated will simply be "skipped" when an allocation occurs in order to find a suitable location for the allocation. For example, 0x30000000-0x30000FFF, being strictly mappable to FCRAM+0 to 0xFFF might not be able to be allocated for the linear heap if something allocated the FCRAM what would be there (noting that other memory regions, e.g. Code Binary and the Regular Heap, aren't strict about physical FCRAM locations), so it might proceed to look at 0x30001000-0x30001FFF, and so on.

CasualPokePlayer: To be clear too, when I say an area is "not mapped" that means that say, if the game were to read from that area on a console, it'd just segfault (data abort exception), although on emulator and for the purpose of peeking memory, 0s would be returned.

ChaseParate: If it helps, this is similar to how the [AMD64/x86-64](https://docs.amd.com/v/u/en-US/40332-PUB_4.08) architecture currently defines/implements a 48-bit virtual address space (see page 584 of the linked document if interested). That allows 256 TiB of addressable memory, even though most users won't have anything close to that much physical memory or swap space.

JoeysRetroHandhelds:
Some comments here if you want them:

- 0x14000000 and 0x30000000 seem to be mutually exclusive — a game picks one based on its kernel version, so one is always zeros during a search, this I believe was changed in firmware 8.0.0-18
- May be worth marking 0x14000000 as RC_MEMORY_TYPE_UNUSED since virtually all achievement targets that firmware
- Should reduce the searchable footprint too

`TEST_PARAMS3(test_memory, RC_CONSOLE_NINTENDO_3DS, 0x40000000, 0x24E00000);`

Should verify memory maps total size and searchable size are both correct.

denzi-gh:
I added a small 3DS Memory Map dump function in Azahar that dumps the VMA layout live while the game runs. It records the vaddr range, soze, permissions, memory state, which 3DS region it falls in and the backing group, backing offset and the alias group.

The Memory Map dumps for the following games can be found under [this forked Azahar Repo](https://github.com/denzi-gh/azahar/tree/bd19874e7957e403bc8eff5370bd963001feb960/docs/3ds_memorydumps). When the forked Azahar repo compiles, you can dump a snapshot of the running game under View -> Debugging -> Dump RA 3DS Memory Map

I tested following games and made sure they are old 3DS titles, newer ones, new 3DS exclusive games, small and big titles:
| Game                      | Linear heap (MiB) | Regular heap (MiB) | mapped total (MiB) | aliasing      | layout  |
|---------------------------|------------------|-------------------|-------------------|---------------|---------|
| Pokémon Ultra Moon        | New 65.5         | 4.4               | 82–84             | code overlays | changes |
| Pokémon Omega Ruby        | Old 46.5         | 14.2              | 72–74             | code + data   | changes |
| New Super Mario Bros. 2   | Old 61.4         | 0.02              | 73                | none          | static  |
| Minecraft New 3DS Ed.     | New 108.0        | 8.3               | 130               | data only     | static  |
| Fire Emblem Warriors      | New 50.0         | 68.75(lol)             | 130               | none          | static  |
| Monster Hunter 4 Ultimate | New 58.4         | 49.5              | 132               | none          | static  |
| Xenoblade Chronicles 3D   | New 77.0         | 41.3              | 132               | none          | static  |

my main takeaways are:

- 0x14000000 (Old) and 0x30000000 (New) are infact mutually exclusive, but should not be marked UNSUED. Omega Ruby uses the Old heap at 0x14000000 and NSMB2 too, so we need to keep both regions in SYSTEM_RAM. Big win nonetheless because Achievement devs can ignore one linear heap entirely everytime.

- the regular heap ranges from 0.02 to 68.75 MiB across the 7 games so it needs to be included as well

- Layouts are mostly static across all games, 2 out of the 7 games (both Pokemon) have Code Overlay swaps. In the Party Menu there is 1 AliasCode overlay, in the Overworld 4 and in Battle 5

- Aliasing is small, for the data sample I have, prefering the lowest virtual adress would be mostly enough without seeing to much copys, can test that further tho

- across all games only 72-132MB actually backed and stayed constant across all dump states, so from roughly 580 MB only 72-132 MB (in this dataset) would need to be searched, the rest reads as 0 and from the top of my head could be filtered out with a "changed value" search pretty reasonable

I also dumped two states from Ultra Moon while in Old 3DS Mode and the virtual layout is 1:1 identical - so I assume Old or New 3DS Mode do not matter while creating Sets.

### in Conclusion:

Changing line 781 to:
```C
{ 0x38000000U, 0x3FFFFFFFU, 0x38000000U, RC_MEMORY_TYPE_UNUSED, "" } /* Extended-FCRAM half of the new linear heap: likely not reachable by any retail game*/
```
and marking this Area UNUSED would shrink the ~750MB Area to ~590MB. And if @Jamiras makes the HARDWARE_CONTROLLER, VIDEO_RAM and REAONLY regions non-searchable the area shrinks to **~451MB of SYSTEM_RAM with 72-132MB that needs to be actually searched**.

Hope this helps and unblocked some barriers for the rollout!

src/rcheevos/consoleinfo.c
denzi-gh: to add to this - like @CasualPokePlayer said, the FCRAM backed SYSTEM_RAM ranges are maximum extents that get mapped on the fly into some part of the 128/256MB FCRAM as allocations happen, with unmapped adresses reading back as 0

from the 7 games I made 3DS memory map dumps from (details for the dumps is in my other comment further down), Monster Hunter Ultimate 4 was the heaviest and the game backed roughly 132 MiB and it all resolves to the one FCRAM chip with no overlapping offsets

this is from the MHU4 [dump](https://github.com/denzi-gh/azahar/blob/bd19874e7957e403bc8eff5370bd963001feb960/docs/3ds_memorydumps/monster_hunter_4_ultimate/in%20title%20screen/redgiant_0004000000126100_001_20260724_004113.csv):
| vaddr    | size     | region        | fcram_offset |
|----------|----------|---------------|--------------|
| 30000000 | 03A63000 | NewLinearHeap | 00000000     |
| 08000000 | 00DF0000 | RegularHeap   | 03A63000     |
| 00100000 | 00CED000 | ProcessImage  | 06F13000     |

The New Liniear Heap segments also should not be marked as VIRTUAL_RAM, because its not just mapping data. The FCRAM offsets also arent exposed anywhere else - Minecraft new 3DS Edition for example stores roughly 108 MB of unique data in there. The only real mirroring I saw were small data pages (in Omega Ruby and Minecraft)

What could be marked as UNUSED is the New Linear Heap Region `38000000 - 3FFFFFFF`. The New 3DS exclusive title Xenoblade Chronicles 3D (and I think the biggist or second biggest game in size) saturates the 124 MiB app region but the linear heap never passes 0x35000000 and Minecraft linear heap tops out at 0x36BF3FFF - thats roughly 20 MiB still missing, This heap range could only be reached by Dev Mode Applications, but to my knowledge never from Retail games. So marking:

`{ 0x38000000U, 0x3FFFFFFFU, 0x38000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "New Linear Heap (New 3DS Exclusive)" } /* New 3DS exclusive space for the newer linear heap */`

UNUSED would reduce the search area by 128 MiB veri niiice

My general findings I found with the dumps are further down, wanted to clarify on this discussion, I hope it helps!

CasualPokePlayer:
A solution to the massive memory map was already done, by simply having the emulator signal to the toolkit that the read failed outright (distinct from reading 0 bytes), effectively setting memory as "unused" on the fly for a given search. This itself would limit searches to the memory that is actually backed.

PR #462, opened 2025-09-11
align Wii memory map to native addresses

Jamiras:
This should hopefully simplify achievement development as RetroAchievements memory addresses should align with native memory addresses.  More importantly, pointers can be followed from each of the memory banks in the same manner instead of requiring different calculations for each bank.

Caveat: Wii does provide two addresses for each of its memory banks. The memory accessible starting at $80000000 is also accessible from $C0000000, and the memory accessible starting at $90000000 is also accessible from $D0000000. If pointers are using the second addresses, they will still have to convert to the first addresses, but the conversion is the same for both banks.


Jamiras:
The [dolphin team has asked](https://discord.com/channels/310192285306454017/1149693430306447380/1415735825873764362) that we maintain the existing mapping for GameCube when expanding to Wii. As such, the memory at $80000000 will be mapped back to $00000000 again. However, padding will still be provided between that memory and the $90000000 memory such that the $90000000 memory will be accessible at $10000000, allowing a mask of 0x1fffffff to map real addresses to RA addresses regardless of whether the pointed-at data is in the GameCube memory region or the Wii expanded memory region.

PR #487, opened 2025-12-07
Mark status code 520 as retryable

adamantike:
This change considers the HTTP status code 520 ("Web Server Returned an Unknown Error") as a retryable error.

CloudFlare documentation reference:
https://developers.cloudflare.com/support/troubleshooting/http-status-codes/cloudflare-5xx-errors/error-520/

### DIFF
--- a/src/rapi/rc_api_common.c
+++ b/src/rapi/rc_api_common.c
@@ -375,6 +375,10 @@ int rc_json_parse_server_response(rc_api_response_t* response, const rc_api_serv
         response->error_message = "Request has timed out.";
         break;
 
+      case 520: /* 520 Unknown Error */
+        response->error_message = "Unknown server error occurred.";
+        break;
+
       case 521: /* 521 Web Server is Down */
       case 523: /* 523 Origin is Unreachable */
         response->error_message = "Could not connect to server.";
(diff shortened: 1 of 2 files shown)


Jamiras:
Can you give an example of something that would generate this error? If the server is crashing, we don't want to keep sending the request that caused the server to crash back to the server to cause it to crash more.

A retryable error is something we expect will recover soon.

adamantike:
Sure, that's a great point. I don't think I can find examples, but I can provide more information about an achievement trigger that failed on my end with error 520 and wasn't retried. That was the reason I started investigating where in RA's code the retryable status codes are defined.

In case the RA team has error tracking or logging enabled, we could try to understand whether this change is useful (e.g. crashes caused by OOM issues that wouldn't be related to the specific request but to current server state).

The unlock request that failed for my user was on achievement [170484](https://retroachievements.org/achievement/170484), which based on RetroArch's automatic screenshot, was triggered at `2025-11-30 20:43:23 UTC`. I got a different achievement (`170482`) two minutes after that, so I think the failed request was just a fluke.

Jamiras:
We do capture exceptions (which includes out of memory errors). I only see two errors reported on 11/30, but neither matches your timestamp. And neither came from API calls.

image

The serialization error occurs a couple times on the following day as well, but there's nothing suggesting an issue with achievement unlocks.

And searching the error logs for "170484" and "adamantike" don't generate any results.

Unfortunately, we only keep about 48 hours of access logs, so I can't try to trace back to your particular interaction. But the only references to 520 I can find in it are content lengths. The 520 error code is happening at the Cloudflare level, so it's unclear what's happening on our end without some sort of way to correlate the events.

PR #497, opened 2026-01-23
return 0 on error in rc_runtime_progress_size()

stenzek:
Otherwise it ends up casting a negative error status to uint32_t.


Jamiras (approved):
The return type was changed from `int` to `uint32_t` in 11.2.

It looks like the only error that can be returned currently is `RC_INSUFFICIENT_BUFFER`, which doesn't apply when calculating the size of the buffer, so this code shouldn't technically ever get hit.

PR #502, opened 2026-03-24
Add offline mode with persistent caching and unlock queue

clankur:
## Summary

- Adds a storage callback interface (`rc_client_set_storage_callbacks`) that hosts provide for persistent file I/O, keeping rcheevos platform-agnostic
- Caches login credentials, game definitions (raw JSON from `r=fetch_game_sets`), and unlock status (from `r=startsession`) to enable fully offline startup
- Implements a persistent binary queue that saves achievement unlocks to disk immediately on trigger, surviving emulator crashes
- On reconnect, flushes queued unlocks via individual `awardachievement` calls with accurate `seconds_since_unlock` timestamps
- Forces all offline unlocks to **softcore only** (client-side enforcement)
- Discards queue entries older than 14 days at flush time
- Fully backward compatible: hosts that don't set storage callbacks get current behavior (all offline features are no-ops)

### New public API
- `rc_client_set_storage_callbacks(client, read_func, write_func)`
- `rc_client_get_offline(client)` — returns non-zero when in offline mode
- `RC_CLIENT_EVENT_OFFLINE_MODE_CHANGED` — event fired on offline/online transitions

### Files
| File | Change |
|------|--------|
| `include/rc_client.h` | Storage callback typedefs, event constant, new API declarations |
| `src/rc_client_internal.h` | Callback fields + offline state flag |
| `src/rc_client.c` | 9 integration points (login, game load, session, award, destroy) |
| `src/rc_client_offline.h` | New internal header for offline module |
| `src/rc_client_offline.c` | Cache + queue implementation (~530 lines) |
| `test/test_rc_client_offline.c` | 18 unit tests for all offline components |

## Test plan
- [x] 18 unit tests pass (CRC32, credential cache, game data cache, queue ops, corruption handling, backward compat)
- [x] Compiles cleanly with `-Wall -Wextra -std=c99`
- [ ] Integration test: online login → cache primed → offline restart → game loads from cache → achievement triggers → queue to disk → online login → queue flushed
- [ ] Verify existing `test_rc_client.c` test suite still passes
- [ ] Test on RetroArch with storage callbacks wired up

## Note
Primarily generated with Claude with some guidance from myself

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/rc_client.h
+++ b/include/rc_client.h
@@ -44,6 +44,18 @@ typedef void (RC_CCONV *rc_client_callback_t)(int result, const char* error_mess
  */
 typedef void (RC_CCONV *rc_client_message_callback_t)(const char* message, const rc_client_t* client);
 
+/**
+ * Callback used to write data to persistent storage.
+ * Returns bytes written, or -1 on error.
+ */
+typedef int (RC_CCONV *rc_client_write_storage_func_t)(const char* filename, const uint8_t* data, uint32_t data_size, rc_client_t* client);
+
+/**
+ * Callback used to read data from persistent storage.
+ * Returns bytes read into buffer, or -1 on error. 0 indicates file not found.
+ */
+typedef int (RC_CCONV *rc_client_read_storage_func_t)(const char* filename, uint8_t* buffer, uint32_t buffer_size, rc_client_t* client);
+
 /*****************************************************************************\
 | Runtime                                                                     |
 \*****************************************************************************/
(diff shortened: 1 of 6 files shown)


Hexadigital:
How does this handle a scenario where:
1. A new achievement set releases
2. The player loads up the achievement set for the first time, to play later
3. The achievement developer realizes that a typo in their logic caused all of the stage checks to be on stage 1, allowing the player to unlock every progression achievement by completing the first stage
4. The player plays the game offline and unlocks all of those achievements erroneously
5. The player comes back online to sync their unlocks

In this scenario, if the player had been online, they would not have been able to unlock all of those broken achievements since they would have gotten the fixed logic.

wescopeland:
Screenshot 2026-03-24 at 7 29 57 AM

`?r=fetch_game_sets` does not exist in the Connect API. The actual endpoint is `achievementsets`. This indicates the code was written without validating against the API.

RAWeb has an [AI policy](https://github.com/RetroAchievements/RAWeb/blob/master/docs/AI_POLICY.md) which explicitly states "You must understand what you are submitting." If this policy were extended org-wide, the disclosure requirement here would be met, but disclosure alone is not sufficient. 1,378 lines in a single commit for a major architectural feature is exactly the kind of contribution that requires its author to deeply understand every line. The non-existent endpoint mentioned in the PR description suggests that level of understanding is not present.

@Hexadigital's concern about stale cached achievement data is also a real design issue that needs to be addressed.

I see two paths forward:
1. Close this, and reopen later as a smaller, incremental series of PRs that can each be reviewed and tested meaningfully, or
2. It remains open with a significantly elevated review bar. There should be a demonstrated working integration against the actual Connect API, the stale data concern should be addressed with a concrete invalidation strategy, and there should be manual testing evidence (ideally in an emulator) for each behavior the PR introduces.

clankur:
Ah sorry @wescopeland I drafted this PR not expecting to get eyes on it. Closing it for now, will probably look into chunking this to align with the past PRs so that its more manageable after also verifying it works E2E. Will try incorporating @Hexadigital's comments as I get it working. 

Thanks for your quick feedback guys!

PR #503, opened 2026-04-05
feat: add native CHD support to cdreader (createcd + createdvd)

lazepeda01:
## Problem

`cdreader.c` has never had native CHD support. When a `.chd` file is passed to the hasher, `cdreader_open_track_iterator` falls through to `cdreader_open_bin_track`, which attempts to read the CHD as a raw binary. This fails because CHD is a compressed container format. Every console attempt prints `Could not open track` and hash generation fails entirely.

This affects all CHD users but is especially visible with PS2 and PSP games compressed using `chdman createdvd`, which is the format chdman now recommends for DVD-based games. These produce a CHD with **no CD track metadata at all**, causing total failure even in clients that implement their own CHD handler but don't handle the DVD format.

> **Note:** RetroArch PR #18040 fixed this within RetroArch's own internal CHD handler. However, that fix only applies to RetroArch itself. rcheevos' default cdreader has never had any CHD support, meaning every other client that calls `rc_hash_init_default_cdreader()` still fails entirely. This PR fixes the root cause at the library level so all clients benefit.

**Related upstream issues:**
- libretro/RetroArch#18005
- libretro/RetroArch#16145

---

## Solution

Add native CHD reading to `cdreader.c` via [libchdr](https://github.com/rtissera/libchdr), added as a submodule at `deps/libchdr`.

The feature is gated behind a `HAVE_CHD` compile flag so existing clients that supply their own CHD callback (e.g. RetroArch) can opt out and keep their own implementation unchanged.

### Two CHD formats are now supported

**`createcd` CHDs**
Contain `CDROM_TRACK_METADATA` / `CDROM_TRACK_METADATA2` entries. We scan them to locate the requested track and configure sector geometry (size, header offset) correctly per track type (`MODE1`, `MODE2`, `AUDIO`). All track selection modes are supported: exact number, `RC_HASH_CDTRACK_FIRST_DATA`, `RC_HASH_CDTRACK_LARGEST`, `RC_HASH_CDTRACK_LAST`.

**`createdvd` CHDs**
Contain no CD track metadata. Detected by `header->unitbytes == 2048` and `Metadata Tag='DVD '`. Synthesised as a single virtual track covering all raw 2048-byte DVD sectors with no header stripping needed. This is the root cause of issues #18005 and #16145.

---

## Implementation notes

- `rc_chd_reader_t` wraps the open `chd_file*` and presents a `rc_hash_filereader_t` interface, so `cdreader_read_sector` and `cdreader_close_track` work completely unchanged.
- Hunk decompression is cached (`buffered_hunk`) to avoid repeated `chd_read` calls when reading across sector boundaries.
- libchdr supports all CHD compression codecs including **Zstandard (zstd)**, which is what modern `chdman` uses by default. Real-world PS2 CHDs downloaded today typically use zstd. Without this fix they fail even in clients that partially handle CHD.
- No changes to `hash_disc.c` or any console-specific hash logic; the fix is entirely in the CD reader layer.
- Clients that already handle CHD (e.g. RetroArch) are unaffected; they override `open_track_iterator` before our code runs.

---

## Real-world verification

Inspecting a real-world downloaded PS2 CHD with `chdman info` confirms the exact format this fix targets:

chdman_test

Tested against 3 PS2 titles, each hashed from both the original ISO and a `createdvd` CHD. All three produce identical hashes:

| Game | Hash |
|---|---|
| Shrek 2 (USA) | `3d9206a51e2ab540c3c0f344cd30520d` |
| Star Wars: Battlefront II (USA) (v2.01) | `ffb7e6298d9386cc197af821e67e58f9` |
| Spider-Man (USA) (v2.01) | `c7b60699f5e993f3fd89a163eda917b0` |

hash_test

---

## Build & test
```bash
# Baseline — no CHD, all existing tests pass
make ARCH=x64 BUILD=c89

# With CHD enabled — all existing tests still pass
make ARCH=x64 BUILD=c89 HAVE_CHD=1
```

All 2720 existing tests pass in both configurations.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "deps/libchdr"]
+	path = deps/libchdr
+	url = https://github.com/rtissera/libchdr.git
--- a/src/rhash/cdreader.c
+++ b/src/rhash/cdreader.c
@@ -749,11 +749,142 @@ static void* cdreader_open_gdi_track(const char* path, uint32_t track, const rc_
   return cdrom;
 }
 
+#ifdef HAVE_CHD
+#include "libchdr/chd.h"
+#include "libchdr/cdrom.h"
+typedef struct rc_chd_reader_t {
+  chd_file* chd;
+  uint32_t hunkbytes;
+  uint8_t* hunk_buffer;
+  uint32_t buffered_hunk;
+  int64_t current_pos;
+  int64_t total_bytes;
+} rc_chd_reader_t;
+#endif /* HAVE_CHD */
+
+#ifdef HAVE_CHD
+static void chd_filereader_seek(void* h, int64_t o, int origin) {
+  rc_chd_reader_t* r = (rc_chd_reader_t*)h;
+  if (origin == SEEK_SET) r->current_pos = o;
+  else if (origin == SEEK_CUR) r->current_pos += o;
+  else r->current_pos = r->total_bytes + o;
+  if (r->current_pos < 0) r->current_pos = 0;
+  if (r->current_pos > r->total_bytes) r->current_pos = r->total_bytes;
+}
+static int64_t chd_filereader_tell(void* h) { return ((rc_chd_reader_t*)h)->current_pos; }
+static size_t chd_filereader_read(void* h, void* buf, size_t n) {
+  rc_chd_reader_t* r = (rc_chd_reader_t*)h;
+  uint8_t* out = (uint8_t*)buf;
+  size_t total = 0;
+  while (n > 0 && r->current_pos < r->total_bytes) {
+    uint32_t hi = (uint32_t)(r->current_pos / r->hunkbytes);
+    uint32_t off = (uint32_t)(r->current_pos % r->hunkbytes);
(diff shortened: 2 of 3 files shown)


Jamiras:
rcheevos should not link directly to libchdr as that will cause conflicts with upstream projects integrating with libchdr and rcheevos. 

Instead, the upstream project should register its own cdreader functions using a custom cd reader (see https://github.com/RetroAchievements/rcheevos/wiki/rc_hash_init_custom_cdreader#remarks).

lazepeda01:
Thanks for the explanation, that makes sense.

For what it's worth, this confirmed that `createdvd` CHDs are identifiable by `header->unitbytes == 2048` with no `CDROM_TRACK_METADATA` entries, and that serving them as a flat 2048-byte sector stream produces hashes identical to the source ISO. This may be useful context for anyone implementing a custom cdreader for CHD support, especially for PS2/PSP games compressed with `chdman createdvd`.

I'll just go ahead and close this PR then. Thanks for the review.

PR #507, opened 2026-04-10
Use _vsnprintf_s() in rc_client_log_message_va()

stenzek:
vsprintf_s() will abort the program if the buffer is too small to contain the message, which can happen with long error messages.

Instead, use _vsnprintf_s() with _TRUNCATE, which matches the non-Windows vsnprintf() behavior.

### DIFF
--- a/src/rc_client.c
+++ b/src/rc_client.c
@@ -240,7 +240,7 @@ static void rc_client_log_message_va(const rc_client_t* client, const char* form
     char buffer[2048];
 
 #ifdef __STDC_SECURE_LIB__
-    vsprintf_s(buffer, sizeof(buffer), format, args);
+    vsnprintf_s(buffer, sizeof(buffer), _TRUNCATE, format, args);
 #elif __STDC_VERSION__ >= 199901L /* vsnprintf requires c99 */
     vsnprintf(buffer, sizeof(buffer), format, args);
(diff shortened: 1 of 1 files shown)


src/rc_client.c
```diff
@@ -240,7 +240,7 @@ static void rc_client_log_message_va(const rc_client_t* client, const char* form
     char buffer[2048];
 
 #ifdef __STDC_SECURE_LIB__
-    vsprintf_s(buffer, sizeof(buffer), format, args);
+    _vsnprintf_s(buffer, sizeof(buffer), _TRUNCATE, format, args);
```
Jamiras: Is there a reason to use `_vsnprintf_s` over `vsnprintf_s`? The [documentation](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/vsnprintf-s-vsnprintf-s-vsnprintf-s-l-vsnwprintf-s-vsnwprintf-s-l?view=msvc-170#remarks) says they're identical:

> vsnprintf_s is identical to _vsnprintf_s and is included for conformance to the ANSI standard.

Jamiras (approved):

src/rc_client.c
stenzek: Nope. I just used it out of habit, will swap it out.

PR #509, opened 2026-04-14
Prevent popup on MeasuredIf activation

KiliLoje:
This prevent the popup that happens because the value goes from being 0ed out by the measuredif to the actual value when all measuredifs become true.

It would allow a whole bunch of control over the pop ups with the logic and we would be able to do things like make the progress tracker pop up appear only once every 10 increments or even crazyier things that I can't even think of, all that while still having the measured value be readable in the retroacheivement menu of the emulator.

The problem would be if some sets rely on this specific pop up but that sounds a bit weird and I don't think any achievements actually do that.

### DIFF
--- a/include/rc_runtime_types.h
+++ b/include/rc_runtime_types.h
@@ -289,6 +289,9 @@ struct rc_trigger_t {
   /* True if the measured value should be displayed as a percentage */
   uint8_t measured_as_percent;
 
+  /* True if all MeasuredIf conditions are true */
+  uint8_t can_measure;
+
   /* True if the trigger has its own rc_memrefs_t */
   uint8_t has_memrefs;
 };
--- a/src/rc_client.c
+++ b/src/rc_client.c
@@ -5834,12 +5834,14 @@ static void rc_client_do_frame_process_achievements(rc_client_t* client, rc_clie
     rc_trigger_t* trigger = achievement->trigger;
     int old_state, new_state;
     uint32_t old_measured_value;
+    uint8_t old_can_measure;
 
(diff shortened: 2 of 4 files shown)


Jamiras:
MeasuredIf is designed to prevent tracking when some precondition is not met, such as counting kills, but only when playing as Bob.

To prevent a Measured tracker from popping up at unexpected times, use PauseIf: https://docs.retroachievements.org/developer-docs/flags/measured.html#using-pause-if-with-measured

KiliLoje:
Alright that makes sense. My bad there.

KiliLoje:
Alright I did some tests, and you can prevent the pop ups with PauseIf. This works with the downside of not having an accurate value in the retroachievement menu, which is the whole point of that PR. The value shown in the achievement list will be the paused value and not the true value

Edit : I understand if you still think that this is not worth it

Jamiras:
PauseIf is meant to prevent garbage from showing up in the popups and the menu when the memory gets temporarily repurposed (such as during a cutscene). 

Preventing the value from being displayed when it is correctly updating may be misleading for players. Using your example of only updating every 10, a player could do the thing 9 times and not know they've done it at all. And if you only want to measure multiples of 10, you can do that with math: "Remember (value) / 10; AddSource {recall} * 10; Measured 0 = (target)".

If you're tracking a big number that you only want to update occassionally, consider using Measured %, which will only pop up when the percentage changes.

And as for "even crazyier things that I can't even think of", we shouldn't implement stuff without an intended use case.

KiliLoje:
Got it. Sorry for insisting.

Have a great day  :)

PR #514, opened 2026-04-30
Adds the extended RAM to the psp memory map

KiliLoje:
The extended RAM from the PSP-2000 and onward is used by a lot of homebrews which are currently not devvable.
This include the minecraft homebrew.

the flag `MEMSIZE` can be set in `param.sfo` :
- to 0 to add an extra 16MB of RAM 
- to 1 to add 32MB

If that flag isn't set by the loaded game, then the system's RAM size won't be upgraded by PPSSPP


Jamiras (approved):
Seems to work both with PPSSPP and RAlibretro. Tested Minecraft (as mentioned) and a retail game that doesn't expose the extra memory.

PR #515, opened 2026-05-05
[WIP] Wii U memory map + .rpx hashing

Loloseiz:
I have mapped the Wii U virtual memory based on Cafe OS specifications and verified the bounds against Cemu's open-source allocator (`coreinit_Memory.cpp`).

* **MEM2 (2GB):** `0x10000000` -> `RC_MEMORY_TYPE_SYSTEM_RAM`
* **MEM1 (32MB):** `0xF4000000` -> `RC_MEMORY_TYPE_VIRTUAL_RAM`
* **Code/Heap (240MB):** `0x01000000` -> `RC_MEMORY_TYPE_READONLY`

I saw Jamiras's previous concerns on the 3DS PR regarding the massive memory dumps, so I marked MEM1 as `VIRTUAL_RAM` and the Code segment as `READONLY` to restrict the default searchable pool strictly to the MEM2 block, not sure if much more can / needs to be done as of now.

I also implemented the ELF parser in `hash_rom.c` to identify and hash `.rpx` executables, along with custom unit tests.

Let me know if I should change something, I tried to follow the baseline set with previous functions and methods, hope this will be useful !


src/rcheevos/consoleinfo.c
```diff
@@ -976,6 +976,17 @@ static const rc_memory_region_t _rc_memory_regions_wii[] = {
...
+/* ===== Wii U ===== */
+/* https://wiiubrew.org/wiki/Memory_map */
+static const rc_memory_region_t _rc_memory_regions_wii_u[] = {
+    { 0x00000000U, 0x00FFFFFFU, 0x00000000U, RC_MEMORY_TYPE_UNUSED, "Unused" },
+    { 0x01000000U, 0x0FFFFFFFU, 0x01000000U, RC_MEMORY_TYPE_READONLY, "Code / Heap" },
+    { 0x10000000U, 0x8FFFFFFFU, 0x10000000U, RC_MEMORY_TYPE_SYSTEM_RAM, "MEM2" },
+    { 0x90000000U, 0xF3FFFFFFU, 0x90000000U, RC_MEMORY_TYPE_UNUSED, "Unused" },
+    { 0xF4000000U, 0xF5FFFFFFU, 0xF4000000U, RC_MEMORY_TYPE_VIRTUAL_RAM, "MEM1" }
+};
+static const rc_memory_regions_t rc_memory_regions_wii_u = { _rc_memory_regions_wii_u, 5 };
```
Jamiras, line 988: > I saw Jamiras's previous concerns on the 3DS PR regarding the massive memory dumps, so I marked MEM1 as VIRTUAL_RAM and the Code segment as READONLY to restrict the default searchable pool strictly to the MEM2 block, not sure if much more can / needs to be done as of now.

Excluding the 32MB MEM1 while still including the 2048MB MEM2 doesn't really seem like it would help much. I've been very reluctant to proceed with the 3DS PR because it's trying to expose 1/3 as much memory as this PR does.

When a user presses "New Search", we ask the core/emulator for *all* the memory expected by the memory map. This gets saved in a snapshot so when the user presses "Filter", we can identify which memory values changed. With a 2GB+ initial snapshot, and large subsets of that for each step of filtering, the host system will very quickly run out of memory before the developer is able to find the data they're looking for.

src/rhash/hash_rom.c
```diff
@@ -424,3 +424,27 @@ int rc_hash_snes(char hash[33], const rc_hash_iterator_t* iterator)
 
   return rc_hash_iterator_buffer(hash, iterator);
 }
+
+int rc_hash_wiiu(char hash[33], const rc_hash_iterator_t* iterator)
```
Jamiras: Conceptually, this should be `rc_hash_wiiu_rpx`. Most users will expect support for `wux` files, and this function won't handle those.

Additionally, `wux` processing should be in `hash_disc.c`, so there shouldn't be a common function that tries to support both.

test/rhash/test_hash_rom.c
```diff
@@ -700,6 +700,70 @@ static void test_hash_scv_cart()
...
+
+  /* Found with TDD */
+  const char* expected_md5 = "2c0b393597ad45952138f79c7180fc88";
+
+  image[0] = 0x7F;
+  image[1] = 'E';
+  image[2] = 'L';
+  image[3] = 'F';
+  image[5] = 0x02;  /* Big Endian */
+  image[18] = 0x14; /* PowerPC */
```
Jamiras, line 717: These comments should be duplicated in the validation logic.

test/rhash/test_hash_rom.c
```diff
@@ -891,6 +955,10 @@ void test_hash_rom(void) {
   /* WASM-4 */
   TEST_PARAMS4(test_hash_full_file, RC_CONSOLE_WASM4, "test.wasm", 33454, "bce38bb5f05622fc7e0e56757059d180");
 
+  /* Wii U */
+  TEST(test_hash_wiiu_rpx);
+  TEST(test_hash_wiiu_invalid);
```
Jamiras: With the implied future support of `wux` files, this should be `test_hash_wiiu_rpx_invalid`.

Jamiras (commented):
I really don't feel like we're in a position to support Wii U at this time (or any time soon).

Loloseiz:
> I really don't feel like we're in a position to support Wii U at this time (or any time soon).

Yeah, I figured by looking at other parts of the project later on. I'll proceed with the requested changes and leave this PR as a Draft/reference. I hope it will be helpful when the team is ready to tackle this generation of consoles :)

src/rhash/hash_rom.c
Loloseiz: > Most users will expect support for wux files, and this function won't handle those.

Fully parsing WUD/WUX inside `hash_disc.c` would require `rcheevos` to implement Wii U AES decryption and manage proprietary keys files, which feels a bit out of scope (from what I understand)

Maybe we could bypass this and fully support `.wux` files without bloating this repo by letting Cemu handle the heavy lifting, since it already decrypts and extracts `main.rpx` into RAM. We could theoretically pass the buffer directly to `rc_hash_generate_from_buffer()` from there.

With that, `rcheevos` would only ever need to know how to hash a `.rpx` payload, no matter what file type was initially passed.

That being said, I understand that it's easier said than done, and may not fit what you envision for the project, but I just wanted to leave that there as an idea for future implementations.

Issue #302, opened 2023-12-17
Address core memory swizzling (somehow)


CasualPokePlayer:
Sega Saturn has a particular issue where memory appears to be byteswapped, and pointers are 4-bytes large, resulting in difficulties for creating achievements.

This is rooted in a core issue of libretro memory handling, which rcheevos ends up expecting. libretro will just obtain pointers to core memory. Memory will proceed to be read, byte by byte, to construct memory values.

In theory, this all should work ok. In practice, some cores utilize a trick known as swizzling. This is so an emulated system with an endianess opposite of the host system can directly do common read/writes (assuming the width for common read/writes is greater than 1 byte) under the host system's endianess. When it comes time to do a 1 byte read, the address will be xored with a constant before being used to index a memory buffer (for 2 byte swizzling, this constant is 1, for 4 byte swizzling, this constant is 3).

This knowledge is even worse when you consider this is only going to possibly occur when the host system's endianess is opposite of the emulate system's endianess. If you change the host system so something which matches the emulated system's endianess, you will now read memory "correctly" as it's no longer being swizzled. Of course, this ends up resulting in a mismatch compared to the swizzled memory, and sets would only be made one way or the other.

In practice of course, this trick is limited to emulated big endian systems, as most host systems will be little endian. Not all emulated big endian systems will utilize this trick anyways, some like the Jaguar and GameCube/Wii have big endian memory, but emulators for those systems do not swizzle memory.

The selection of consoles which this trick actually occurs is also very limited. Sega Mega Drive / CD / 32X all utilize this (and of course those are more or less all under the same emulators), Sega Saturn also utilizes this. N64 utilizes this, but instead of simple byteswapping, it does the swizzling by 4 bytes (so ^ 3 instead of ^ 1 for 1 byte reads). (There could be some other systems I'm missing here)

Also in practice, the issues of a big endian host system typically won't actually occur. This will mean the user is running RetroArch on a Wii, Wii U, or old PowerPC Mac, and achievement hunting, and probably doing so with the Genesis Plus GX core. A combination of factors so unlikely it apparently has never actually resulted in some ticket being filed, as I'm told.

As far as I can tell, the libretro API does not offer any way to know if memory is being swizzled or not. It seems to have some kind of flags for retro_memory_map's, so perhaps it'd need a flag added?

[...]

PR #517, opened 2026-05-18
hash: content-hash .neo (Geolith Neo Geo cart) files

codingncaffeine:
## Summary

Registers the `.neo` file extension in the hash-iterator console table under `RC_CONSOLE_ARCADE`, and adds a dedicated hashing function for it: `.neo` files contain the actual ROM data, so they are content-hashed rather than using the arcade filename hash.

## Hashing

The first 4096 bytes of a `.neo` file are a header — magic number, ROM section sizes, then metadata text fields (name, manufacturer, genre). The text fields vary between conversion tools, so the header can't participate in the hash (confirmed with the Geolith author). The decrypted, concatenated ROM data starts at byte 4096 and is deterministic for a given romset, so the hash covers everything from byte 4096 to EOF (capped at `MAX_BUFFER_SIZE`, consistent with `rc_hash_whole_file`). The `NEO\1` magic is validated before hashing.

Other arcade content (`.zip`, `.7z`) is unaffected and keeps the filename hash.

## Background

The [Geolith](https://github.com/libretro/geolith-libretro) libretro core loads Neo Geo cartridges from a self-contained `.neo` file format (one game per file). Geolith now also validates ROMs internally against a checksum database of the canonical set, and that set can be used to seed RA-side hashes — the Geolith author has offered to supply the list.

## Tests

- `test_hash_neo` — hash of a `.neo` file equals the hash of its ROM payload alone
- `test_hash_neo_header_variants` — two files with identical ROM data but different header text fields produce the same hash
- `test_hash_neo_bad_magic` — files without the `NEO\1` magic do not hash

Full suite passes (2733/2733). Branch rebased on current develop.

### DIFF
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -543,6 +543,11 @@ static int rc_hash_from_buffer(char hash[33], uint32_t console_id, const rc_hash
       return rc_hash_buffer(hash, iterator->buffer, iterator->buffer_size, iterator);
 
 #ifndef RC_HASH_NO_ROM
+    case RC_CONSOLE_ARCADE:
+      /* .neo (Geolith Neo Geo cart) files carry the ROM data; other arcade
+       * formats are archives, which aren't hashed from a buffer. */
+      return rc_hash_neogeo_cart(hash, iterator);
+
     case RC_CONSOLE_ARDUBOY:
       return rc_hash_arduboy(hash, iterator);
 
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -1230,6 +1230,7 @@ static const rc_hash_iterator_ext_handler_entry_t rc_hash_iterator_ext_handlers[
   { "n64", rc_hash_initialize_iterator_single, RC_CONSOLE_NINTENDO_64 },
   { "ndd", rc_hash_initialize_iterator_single, RC_CONSOLE_NINTENDO_64 },
   { "nds", rc_hash_initialize_iterator_single, RC_CONSOLE_NINTENDO_DS }, /* handles both DS and DSi */
+  { "neo", rc_hash_initialize_iterator_single, RC_CONSOLE_ARCADE }, /* Geolith Neo Geo cart format */
   { "nes", rc_hash_initialize_iterator_single, RC_CONSOLE_NINTENDO },
   { "ngc", rc_hash_initialize_iterator_single, RC_CONSOLE_NEOGEO_POCKET },
   { "nib", rc_hash_initialize_iterator_nib, 0 },
--- a/src/rhash/hash.c
+++ b/src/rhash/hash.c
@@ -862,6 +862,11 @@ static int rc_hash_from_file(char hash[33], uint32_t console_id, const rc_hash_i
 
 #ifndef RC_HASH_NO_ROM
     case RC_CONSOLE_ARCADE:
+      /* .neo files (Geolith Neo Geo cart format) contain the actual ROM data,
+       * so are content-hashed. Everything else (.zip/.7z) hashes by filename. */
+      if (rc_path_compare_extension(path, "neo"))
+        return rc_hash_neogeo_cart(hash, iterator);
+
       return rc_hash_arcade(hash, iterator);
 
     case RC_CONSOLE_ARDUBOY:
--- a/src/rhash/hash_rom.c
+++ b/src/rhash/hash_rom.c
@@ -359,22 +359,25 @@ int rc_hash_neogeo_cart(char hash[33], const rc_hash_iterator_t* iterator)
   md5_init(&md5);
 
   buffer = (uint8_t*)malloc(chunk_size);
-  if (buffer) {
-    rc_file_seek(iterator, file_handle, (int64_t)header_size, SEEK_SET);
-    while (remaining >= chunk_size) {
-      rc_file_read(iterator, file_handle, buffer, (int)chunk_size);
-      md5_append(&md5, buffer, (int)chunk_size);
-      remaining -= chunk_size;
-    }
-
-    if (remaining > 0) {
-      rc_file_read(iterator, file_handle, buffer, (int)remaining);
-      md5_append(&md5, buffer, (int)remaining);
-    }
-
-    free(buffer);
-    result = rc_hash_finalize(iterator, &md5, hash);
+  if (!buffer) {
+    rc_file_close(iterator, file_handle);
+    return rc_hash_iterator_error(iterator, "Could not allocate temporary buffer");
   }
+
+  rc_file_seek(iterator, file_handle, (int64_t)header_size, SEEK_SET);
+  while (remaining >= chunk_size) {
+    rc_file_read(iterator, file_handle, buffer, (int)chunk_size);
+    md5_append(&md5, buffer, (int)chunk_size);
+    remaining -= chunk_size;
+  }
+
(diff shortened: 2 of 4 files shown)


src/rhash/hash.c
```diff
@@ -1230,6 +1230,7 @@ static const rc_hash_iterator_ext_handler_entry_t rc_hash_iterator_ext_handlers[
   { "n64", rc_hash_initialize_iterator_single, RC_CONSOLE_NINTENDO_64 },
   { "ndd", rc_hash_initialize_iterator_single, RC_CONSOLE_NINTENDO_64 },
   { "nds", rc_hash_initialize_iterator_single, RC_CONSOLE_NINTENDO_DS }, /* handles both DS and DSi */
+  { "neo", rc_hash_initialize_iterator_single, RC_CONSOLE_ARCADE }, /* Geolith Neo Geo cart format */
```
Jamiras, line 1243: Please also add a test for this. 

Probably just a copy/paste of this one:

https://github.com/RetroAchievements/rcheevos/blob/8b0188fe69e9b125e20cbdb45d3e43f53f98c590/test/rhash/test_hash_rom.c#L709

Jamiras (commented):
I realize this is the most straight-forward way to support the .neo file format, but the reason the arcade files are just doing filename hashes is because rcheevos doesn't know how to unzip a file to generate the actual hash and different compression tools (and even different versions fo the same tools or with different flags) generate different zip files, so the filename was all we could reliably check.

We currently heavily rely on the emulator/core to validate the ROMs, as only matching the filename does not ensure the contents are unmodified. Does geolith have a similar internal check? What's to prevent a user from modifying one of the ROMs to cheat?

Relying on the emulator/core for this has two major downside though: If the emulator/core switches to a new set of ROMs for the same game, we'll still load achievements while the emulator won't load the game. And, though very unlikely, the new ROM set might not be compatible with the achievements, so if the user gets the updated ROM set, they might still not be able to earn the achievements.

To address this, we've talked about (but not really persued) hashing the table-of-contents information from within the zip file (as it contains the CRCs of each file) instead of just the hashname. We wouldn't need a full decompressor implementation to pull that data out of the zip file. That information clearly wouldn't be in a .neo file, so if we did ever make that change, we'd have to ensure the .neo format was still handled differently.

Since .neo is an uncompressed file format, we could do full file hashes to ensure the user is using the proper files. Is the header guaranteed to be the same for every "clean dump", or can it vary depending on uninitialized fields, or other versioning information?

Whether we strip the header or not, having separate hashes for the .neo format would require generating and registering the hash for every supported .neo archive. But if Geolith doesn't do internal validation/enforcement, we (RetroAchievements) may require that.

carmiker:
Geolith does no internal validation outside of ensuring the NEO header is correct. This was a design choice to make life easier for anyone who wishes to load homebrew, and to avoid having to maintain a list of approved ROMs. I designed this emulator to act like a console emulator (and it also defaults to AES mode), where you can load any game without developer intervention, as long as it's in the correct format.

The canonical supported set is called "LunaGarlic", those files are reliable and could be used to validate the game has not been modified. Although, if there is going to be a list of approved hashes, it might be worth just calling this a different system than "arcade" at the same time. If this is a viable path, and "neogeo" is created as its own system, then perhaps the existing FB Neo achievements can be copied over, and memory addresses modified so that I don't have to keep a byteswapped copy of main RAM to adhere to FB Neo's choice of byte ordering. This issue addresses the byte ordering question: https://github.com/RetroAchievements/rcheevos/issues/302 (edit: I just found out libretro has a flag for this but requires memory descriptors, not just raw access to the memory via retro_get_memory_*).

(I am reading "But if Geolith doesn't do internal validation/enforcement, we (RetroAchievements) may require that." to mean that RetroAchievements would maintain the list of hashes, and I can supply the list with associated MAME-style filename in whatever format you want)

carmiker:
Ok, I did a bit of learning and discovered the `RETRO_MEMDESC_BIGENDIAN` flag in libretro and how memory descriptors work.

I added a game database (MD5 checksums) to Geolith to ensure a ROM is "verified" before activating the memory descriptors that allow achievements to work. Currently if I rename a .neo to .zip (without actually compressing), everything seems to work. These checksums can be used on the RetroAchievements side to verify games in the future, but also serve to ensure hacked games won't be used for achievements on the client side, right now.

Jamiras:
Sorry for the delay in responding here, but I was trying to have some decisions made internally.

We've decided that we would rather have `.neo` files hashed normally instead of using the filename logic like we do for arcade zip files.

> Since .neo is an uncompressed file format, we could do full file hashes to ensure the user is using the proper files. Is the header guaranteed to be the same for every "clean dump", or can it vary depending on uninitialized fields, or other versioning information?

> Whether we strip the header or not, having separate hashes for the .neo format would require generating and registering the hash for every supported .neo archive.

I'll leave it up to you to decide whether or not the header should be included in the hash. If it's guaranteed to be the same for every conversion of the same set of files, then you can just use `rc_hash_whole_file`, perhaps like this:
```
    case RC_CONSOLE_ARCADE:
      if (rc_path_compare_extension(path, "neo"))
        return rc_hash_whole_file(hash, iterator);

      return rc_hash_arcade(hash, iterator);
```

If not, you'll want a special hashing function to strip the header - there's several already in hash_rom.c that you could look at to get started.

Once the hashing code is committed, we'll add the hashes to our database.

codingncaffeine:
Reworked per your suggestion — .neo files are now content-hashed. Confirmed with carmiker that the header text fields vary by conversion tool but the ROM data always starts at byte 4096, so the hash skips the header (there's a test asserting two files with different headers but identical ROM data hash the same). Once merged, he can supply checksums from the canonical set for the database side.

src/rhash/hash_rom.c
```diff
@@ -297,6 +297,89 @@ int rc_hash_n64(char hash[33], const rc_hash_iterator_t* iterator)
...
+  else {
+    rc_hash_iterator_verbose_formatted(iterator, "Hashing %s (%u bytes after 4096 byte header)",
+      rc_path_get_filename(iterator->path), (unsigned)size);
+    remaining = (size_t)size;
+  }
+
+  md5_init(&md5);
+
+  buffer = (uint8_t*)malloc(chunk_size);
+  if (buffer) {
```
Jamiras: If the malloc fails, you should return an error.
https://github.com/RetroAchievements/rcheevos/blob/9ade739c173a2ee0b14ef887726c2ad92cbd54e1/src/rhash/hash_rom.c#L226-L230

src/rhash/hash_rom.c
```diff
@@ -297,6 +297,89 @@ int rc_hash_n64(char hash[33], const rc_hash_iterator_t* iterator)
...
+  const size_t chunk_size = 65536;
+  md5_state_t md5;
+  uint8_t* buffer;
+  uint8_t header[4];
+  int64_t size;
+  void* file_handle;
+  size_t remaining;
+  int result = 0;
+
+  if (iterator->buffer) {
```
Jamiras, line 320: If you're going to support buffered hashing, you should add a case statement to `rc_hash_from_buffer`. It's probably safe to assume zip files won't be buffered.

```
#ifndef RC_HASH_NO_ROM
    case RC_CONSOLE_ARCADE:
      return rc_hash_neogeo_cart(hash, iterator);

    case RC_CONSOLE_ARDUBOY:
```

Jamiras (approved):

PR #519, opened 2026-05-24
Fix several buffer bounds errors

Jamiras:
for unexpected inputs. Also addresses some double frees and potential leaks for certain error cases.


src/rhash/hash_disc.c
```diff
@@ -1291,6 +1291,9 @@ static int rc_hash_wiiware(md5_state_t* md5, const rc_hash_iterator_t* iterator,
 
     /* Hash content */
     buffer = (uint8_t*)malloc(buffer_size);
+    if (!buffer)
+      return rc_hash_iterator_error(iterator, "Could not allocated content buffer");
```
Pesa:
```suggestion
      return rc_hash_iterator_error(iterator, "Could not allocate content buffer");
```

PR #520, opened 2026-05-26
Add Xbox memory map

Burrrrris:
Adds memory region information for Xbox (`RC_CONSOLE_XBOX`).

Retail Xbox exposes 64 MiB of physical system RAM from `0x00000000` through `0x03FFFFFF`.

Reference:
https://xboxdevwiki.net/Memory


Jamiras (commented):
Please add a test case to test_consoleinfo

Jamiras (approved):

PR #521, opened 2026-06-07
Parse unlock times as 64-bit integers

stenzek:
Timestamps for unlocks were being silently truncated to 32-bit. This would be more alarming if it wasn't _unsigned_ 32-bit, since that doesn't overflow until 2106. But IMO it's better to be consistent across the board.


src/rapi/rc_api_user.c
Jamiras, line 1: There are some similar constructs in `rc_api_info.c` and `rc_api_runtime.c` as well as one on line 360 that should be converted too.

test/rapi/test_rc_api_common.c
```diff
@@ -492,6 +492,67 @@ static void test_json_get_required_bool() {
...
+  rc_json_field_t field;
+  char buffer[64];
+  time_t value = 2;
+  snprintf(buffer, sizeof(buffer), "{\"Test\":%s}", input);
+
+  assert_json_parse_response(&response, &field, buffer, RC_OK);
+
+  if (expected_result) {
+    ASSERT_TRUE(rc_json_get_timet(&value, &field, "Test"));
+    ASSERT_COMPARE(value, ==, expected, long long, "%lld");
```
Jamiras: `%lld` doesn't exist in c89, so the c89 build will likely fail. I don't remember if it's a compile time or runtime error, but there is an `ASSERT_NUM64_EQUALS` macro that is currently restricted to 32-bit values because of it.

https://github.com/RetroAchievements/rcheevos/blob/9ade739c173a2ee0b14ef887726c2ad92cbd54e1/test/test_framework.h#L179-L180

We should probably have an `ASSERT_TIMET_EQUALS` and update all the references to the fields being touched to use it.

src/rapi/rc_api_user.c
stenzek: How did I miss these.... thanks!

test/rapi/test_rc_api_common.c
stenzek: Added `ASSERT_TIMET_EQUALS` as well as implemented the macro for C99.

test/rapi/test_rc_api_common.c
```diff
@@ -889,10 +950,22 @@ void test_rapi_common(void) {
...
+  TEST_PARAMS3(test_json_get_timet, "3.14159", 3, 1);
+  TEST(test_json_get_optional_timet);
+  TEST(test_json_get_required_timet);
+
   /* rc_json_get_datetime */
   TEST_PARAMS2(test_json_get_datetime, "", -1);
-  TEST_PARAMS2(test_json_get_datetime, "2015-01-01 08:15:00", 1420100100);
-  TEST_PARAMS2(test_json_get_datetime, "2016-02-29 20:01:47", 1456776107);
+  TEST_PARAMS2(test_json_get_datetime, "2015-01-01 08:15:00", (time_t)1420100100LL);
+  TEST_PARAMS2(test_json_get_datetime, "2016-02-29 20:01:47", (time_t)1456776107LL);
```
Jamiras, line 968: Do we really need the `LL` here? (Also 517/518 and 957/958).

I would expect this to fail in the c89 build, but it doesn't:
```
cc -Wall -Werror -Wno-long-long -D_LARGEFILE64_SOURCE -D_FILE_OFFSET_BITS=64 -m32 -O3 -std=c89 -DRC_CLIENT_SUPPORTS_HASH -I../include -I../src/rcheevos -I. -c rapi/test_rc_api_common.c -o rapi/test_rc_api_common.o
```

Jamiras: Apparently, `-Wno-long-long` is only enforced when the `-pedantic` flag is passed, which we do for the main code, but not the test code. Doing so in the test code breaks the `TEST_` macros because `__func__` is not defined.

I guess it's probably fine to leave these here as long as the main code is still c89 safe.

Jamiras (approved):

PR #527, opened 2026-07-13
report warning if same condition exists with both delta and non-delta

Jamiras:
Generates a warning if the user has:
```
   Mem   8-bit 0x1234 = 5
   Delta 8-bit 0x1234 = 5
```

The user most likely meant to make the second condition `Delta 0x1234 != 5` or `Delta 0x1234 = 4`

### DIFF
--- a/src/rcheevos/rc_validate.c
+++ b/src/rcheevos/rc_validate.c
@@ -38,6 +38,7 @@ enum
   /* warnings about potential logic errors */
   RC_VALIDATION_ERR_COMPARING_DIFFERENT_MEMORY_SIZES,
   RC_VALIDATION_ERR_MASK_RESULT_ALWAYS_ZERO,
+  RC_VALIDATION_ERR_SAME_COMPARISON_ACROSS_FRAMES,
 
   /* warnings that some areas of memory should be avoided */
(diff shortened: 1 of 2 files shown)


redwizard42:
Probably reasonable warning, but sometimes I do want to make sure it's been a specific value for at least two frames. But more usually I just need it to b stable for two frames and do delta = mem.

Any plays to allow us to see all warnings (or does that exist)?

Jamiras:
I have no plans to show all errors. Multiple errors could be the result of the same bad code. #475 modified the logic to bubble up the most severe error.

Additionally, there are different tiers of errors. Parse errors happen first (multiple measured, modifying operators on conditional lines, using trigger in leaderboard values). Then the logic validation happens. And finally, the note validation logic happens.

I do have plans to allow a user to suppress a warning (assuming the potential error was intentional - like using a 16-bit read to check two neighboring 8-bit values and getting a note size warning), but the naive implementation of that would be to just hide the warning without generating the next warning.

Is there a particular case you have in mind?

redwizard42:
I only I know I've done this at least once and have a small fear that it'd hide a more relevant earning.

It's probably such a niche case though. Mostly just thinking aloud about a possible pitfall.

PR #534, opened 2026-07-29
rc_hash_arcade: handle files without extensions

### LINKED ISSUE #530: rc_hash_arcade drops the last character when the path has no extension
# rc_hash_arcade drops the last character when the path has no extension

`rc_hash_arcade()` computes the name length by subtracting the extension
pointer, assuming a `.` separator is always there:

```c
/* src/rhash/hash_rom.c */
const char* filename = rc_path_get_filename(iterator->path);
const char* ext = rc_path_get_extension(filename);
size_t filename_length = ext - filename - 1;
```

`rc_path_get_extension()` returns `path + strlen(path)` when there is no dot,
so for an extensionless path the `- 1` eats the last character instead of a
separator.

## Reproducing

Compiling `rc_path_get_extension()` verbatim and applying that same expression:

| Path | filename_length | Hashed string |
|---|---|---|
| `mslug2` | 5 | `mslug` |
| `garou` | 4 | `garo` |
| `cyberlip` | 7 | `cyberli` |
| `sonicwi2.neo` | 8 | `sonicwi2` ✅ |
| `mslug2.zip` | 6 | `mslug2` ✅ |

## Impact

Hit on MiSTer's Neo Geo core with an unzipped Darksoft set, where each romset
is a folder and the path carries no extension.

Usually it just fails: `garo` and `cyberli` are not romsets, so the server
answers 404 and no achievements load.

`mslug` is a romset, though. Loading Metal Slug 2 identifies **Metal Slug 1**,
and unlocks are credited to it. Confirmed on hardware: Metal Slug 2 running,
`b43c8b4ec999588c04dad79bb8bcc745` sent — that is `md5("mslug")`, while the
correct hash for the loaded game is `b1475b391ad385c16feae3a2bfbfc2a0`.

## Suggested fix

Only subtract the separator when one is present:

```c
size_t filename_length = (ext > filename && ext[-1] == '.')
                       ? (size_t)(ext - filename - 1)
                       : strlen(filename);
```

Same result for every path that has an extension.

christiannicola:
Fixes https://github.com/RetroAchievements/rcheevos/issues/530

### DIFF
--- a/src/rhash/hash_rom.c
+++ b/src/rhash/hash_rom.c
@@ -35,7 +35,7 @@ int rc_hash_arcade(char hash[33], const rc_hash_iterator_t* iterator)
   const char* filename = rc_path_get_filename(iterator->path);
   const char* ext = rc_path_get_extension(filename);
   char buffer[128]; /* realistically, this should never need more than ~32 characters */
-  size_t filename_length = ext - filename - 1;
+  size_t filename_length = (*ext) ? (size_t)(ext - filename - 1) : strlen(filename);
 
   /* fbneo supports loading subsystems by using specific folder names.
    * if one is found, include it in the hash.
--- a/src/rhash/hash_rom.c
+++ b/src/rhash/hash_rom.c
@@ -35,7 +35,18 @@ int rc_hash_arcade(char hash[33], const rc_hash_iterator_t* iterator)
   const char* filename = rc_path_get_filename(iterator->path);
   const char* ext = rc_path_get_extension(filename);
   char buffer[128]; /* realistically, this should never need more than ~32 characters */
-  size_t filename_length = (*ext) ? (size_t)(ext - filename - 1) : strlen(filename);
+  size_t filename_length;
+
+  if (!*ext) {
+    /* some emulators will receive a path to the extracted romset instead of a compressed romset.
+     * for those emulators, `iterator->path` will be a folder, and therefore won't have an extension to ignore.
+     * we can assume the folder name is the same as the romset archive name and try to hash it.
+     * since we're not actually hashing the contents of the archive, this is safe to do.
+     * */
+    return rc_hash_buffer(hash, (uint8_t*)filename, strlen(filename), iterator);
+  }
+
+  filename_length = ext - filename - 1;
 
   /* fbneo supports loading subsystems by using specific folder names.
    * if one is found, include it in the hash.
--- a/test/rhash/test_hash_rom.c
+++ b/test/rhash/test_hash_rom.c
@@ -816,12 +816,19 @@ void test_hash_rom(void) {
   TEST_PARAMS2(test_hash_arcade, "game.7z", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "game", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "/game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "/game", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "\\game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "\\game", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "roms\\game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "roms\\game", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "C:\\roms\\game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "C:\\roms\\game", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "/home/user/roms/game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "/home/user/roms/game", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "/home/user/games/game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "/home/user/games/game", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "/home/user/roms/game.7z", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "/home/user/roms/game", "c8d46d341bea4fd5bff866a65ff8aea9");
 
   TEST_PARAMS2(test_hash_arcade, "/home/user/nes_game.zip", "9b7aad36b365712fc93728088de4c209");
   TEST_PARAMS2(test_hash_arcade, "/home/user/nes/game.zip", "9b7aad36b365712fc93728088de4c209");
--- a/test/rhash/test_hash_rom.c
+++ b/test/rhash/test_hash_rom.c
@@ -825,10 +825,15 @@ void test_hash_rom(void) {
   TEST_PARAMS2(test_hash_arcade, "/home/user/nes_game.zip", "9b7aad36b365712fc93728088de4c209");
   TEST_PARAMS2(test_hash_arcade, "/home/user/nes/game.zip", "9b7aad36b365712fc93728088de4c209");
   TEST_PARAMS2(test_hash_arcade, "C:\\roms\\nes\\game.zip", "9b7aad36b365712fc93728088de4c209");
+  TEST_PARAMS2(test_hash_arcade, "C:\\roms\\nes\\game", "9b7aad36b365712fc93728088de4c209");
   TEST_PARAMS2(test_hash_arcade, "C:\\roms\\NES\\game.zip", "9b7aad36b365712fc93728088de4c209");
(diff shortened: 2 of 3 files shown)


src/rhash/hash_rom.c
```diff
@@ -35,7 +35,7 @@ int rc_hash_arcade(char hash[33], const rc_hash_iterator_t* iterator)
   const char* filename = rc_path_get_filename(iterator->path);
   const char* ext = rc_path_get_extension(filename);
   char buffer[128]; /* realistically, this should never need more than ~32 characters */
-  size_t filename_length = ext - filename - 1;
+  size_t filename_length = (*ext) ? (size_t)(ext - filename - 1) : strlen(filename);
```
Jamiras: As this is specific to the `rc_hash_arcade` function, it's probably safe to do. But `iterator->path` is expected to point to a file not a directory.

We've discussed the possibility of extracting the table of contents from the zip file for a more robust arcade hash. That would be incompatible with this solution. But as long as we know this is something that should be supported, we could reconstruct the table of contents from the files in the folder. 

Please add a comment along the lines of:
> Some emulators will receive a path to the extracted romset instead of a compressed romset. For those emulators, iterator->path will be a folder, and therefore won't have an extension to ignore. We can assume the folder name is the same as the romset archive name and try to hash it. Since we're not actually hashing the contents of the archive, this is safe to do.

test/rhash/test_hash_rom.c
```diff
@@ -825,10 +825,15 @@ void test_hash_rom(void) {
...
   TEST_PARAMS2(test_hash_arcade, "C:\\roms\\nes\\game.zip", "9b7aad36b365712fc93728088de4c209");
+  TEST_PARAMS2(test_hash_arcade, "C:\\roms\\nes\\game", "9b7aad36b365712fc93728088de4c209");
   TEST_PARAMS2(test_hash_arcade, "C:\\roms\\NES\\game.zip", "9b7aad36b365712fc93728088de4c209");
+  TEST_PARAMS2(test_hash_arcade, "C:\\roms\\NES\\game", "9b7aad36b365712fc93728088de4c209");
   TEST_PARAMS2(test_hash_arcade, "nes\\game.zip", "9b7aad36b365712fc93728088de4c209");
+  TEST_PARAMS2(test_hash_arcade, "nes\\game", "9b7aad36b365712fc93728088de4c209");
   TEST_PARAMS2(test_hash_arcade, "/home/user/snes/game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "/home/user/snes/game", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "/home/user/nes2/game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "/home/user/nes2/game", "c8d46d341bea4fd5bff866a65ff8aea9");
```
Jamiras: These tests are all related to the fbneo subsystem identification and don't really reflect that functionality you're targetting. Your tests should be in the above `game.zip`/`game.7z` tests.

In fact, I might even go so far as to prevent the logic from executing if an extension is not found.
```
  const char* ext = rc_path_get_extension(filename);
  char buffer[128]; /* realistically, this should never need more than ~32 characters */
  size_t filename_length;

  if (!*ext) {
    /* massive comment goes here */
    return rc_hash_buffer(hash, (uint8_t*)filename, strlen(filename), iterator);
  }

  filename_length = ext - filename - 1;
```

src/rhash/hash_rom.c
christiannicola: Sorry for being late to respond, Github did not notify me that the code has been reviewed.
I updated the code with your suggestions, thanks!

test/rhash/test_hash_rom.c
christiannicola: Thanks for the review, I pushed a commit with your suggestions.

test/rhash/test_hash_rom.c
```diff
@@ -814,6 +814,7 @@ void test_hash_rom(void) {
   /* Arcade */
   TEST_PARAMS2(test_hash_arcade, "game.zip", "c8d46d341bea4fd5bff866a65ff8aea9");
   TEST_PARAMS2(test_hash_arcade, "game.7z", "c8d46d341bea4fd5bff866a65ff8aea9");
+  TEST_PARAMS2(test_hash_arcade, "game", "c8d46d341bea4fd5bff866a65ff8aea9");
```
Jamiras: Since this is replicating pathing behavior, I suggest having tests with full paths:
```
"C:\\roms\\game"
"/home/user/roms/game"
```

Jamiras: Also, the test is currently failing because `rc_hash_initialize_iterator_from_path` can't guess that it should use the `RC_CONSOLE_ARCADE` iterator.

That _may_ be a problem when integrating this upstream - if the console isn't explicitly provided, the hasher won't know which hash algorithm to apply.

christiannicola: Interesting. The tests pass if we add an empty extension with the `RC_CONSOLE_ARCADE` hasher to `rc_hash_iterator_ext_handlers`, but then the `test_hash_file_without_ext` fails because we try to parse an NES rom as a NeoGeo cartridge. 

Right now the code defaults to full file hashing if no handler is found when looking at the file path since there is no handler declared for empty extensions. If we change that by specifying the arcade hasher for empty extensions, then the single test expecting a full file hash for a NES ROM without an extension fails. In theory this change would make the test invalid, however I would need some guidance on if that is still correct behavior.

Maybe we should check for a NeoGeo header before the full file hashing function is used?

Jamiras: We definitely don't want files without extensions falling back to just hashing the filename.

And we can't really check for a NeoGeo header when all we have if a path to a folder.

I'm leaning towards not supporting the discovery feature if an extension is not provided. Do you know if MiSTER is specifying the system when calling the hasher, or is it using the discovery logic?

If we had to support the behavior, it could be possible by registering a handler for "no extension" that returns either `RC_CONSOLE_ARCADE` or `RC_CONSOLE_GAMEBOY` (for full file hash) depending on whether the path is a file or directory. Unfortunately, the virtual file system interface doesn't have an "is directory" function because this wasn't an expected scenario. Maybe it could leverage `rc_file_size` to determine if it's an actual file.

This will make the tests work, but I have no idea if `rc_file_size` will reliably return 0 or a negative value for a directory across all OSes/virtual file system implementations.

```
static void rc_hash_initialize_iterator_no_ext(rc_hash_iterator_t* iterator, int data) {
  (void)data;

  if (iterator->path && rc_file_size(iterator, iterator->path) > 0) {
    /* actual file, try a full file hash */
    iterator->consoles[0] = RC_CONSOLE_GAMEBOY;
  }
  else {
    /* not a file. assume it's a directory */
    iterator->consoles[0] = RC_CONSOLE_ARCADE;
  }
}
```

christiannicola: The MiSTer integration always provides the correct system id when generating the hash.

I would also agree that its risky to support no extensions in the discovery logic, but that is the reason why that one test is failing - `rc_hash_initialize_iterator_from_path` defaults to the whole file hash because we don't find a handler match.

Another solution would be to check for the extension in `rc_hash_initialize_iterator_from_path` - if we have no extension at all then we can fall back to checking for a file size and use the appropriate system id.

I've pushed a commit with this change implemented, hopefully the CI pipelines test enough platforms to see if this change is safe.

Jamiras: It looks like what you've committed is just inlining the handler for "no extension". How is that better than having a handler registered in the list?

If you agree that it's better not to support "no extension" in the discovery logic, the solution would be to modify the test to call a copy of `test_hash_arcade` that matches the hash for the `rc_hash_generate_from_file` behavior, and looks for no match for the `rc_hash_iterate` behavior with a comment explaining why we chose not to support the discovery functionality for these cases.

christiannicola: I chose to inline the handler for the no extension case because the handler table matches single systems; the code here still checks if we're dealing with a file or a directory. Since we're only checking this table in one place (`rc_hash_initialize_iterator_from_path`), I thought this change was fine. We also avoid some allocations doing the search as well since we know that we don't have an extension that we need to search for anyway.

Just to clarify that I understood correctly - we should remove the last two commits and simply add tests that state show we don't support the no-extension case in the discovery logic (`rc_hash_initialize_iterator_from_path` without extension)?

I don't think I'm qualified to call any shots here, so it's much appreciated if you guide me into the right direction here.

Jamiras: > the handler table matches single systems

That's not true. Look at the "bin" or "dsk" handlers. There's just a convienence helper for single systems.

> We also avoid some allocations doing the search as well since we know that we don't have an extension that we need to search for anyway.

That's also not true. Because rcheevos is c89 compatible, all of the stack space is allocated at the top of the function. All you really avoid is a zeroing out of 24 bytes, the lowercasing loop (which will immediately break out as the string is empty), and the bsearch.

> Just to clarify that I understood correctly - we should remove the last two commits and simply add tests that state show we don't support the no-extension case in the discovery logic (rc_hash_initialize_iterator_from_path without extension)?

Yes. As long as we don't have to worry about the `rc_hash_iterate` path (system discovery), then the only code change should be self-contained within `rc_hash_arcade`, and you'll just need alternate test function for the non-file tests.

test/rhash/test_hash_rom.c
```diff
@@ -38,6 +38,28 @@ static void test_hash_arcade(const char* path, const char* expected_md5)
   ASSERT_STR_EQUALS(hash_iterator, expected_md5);
 }
 
+static void test_hash_arcade_no_system_match(const char* path, const char* expected_md5)
```
Jamiras: Suggest renaming function to `test_hash_arcade_from_directory` and adding a comment to the `ASSERT_NUM_EQUALS(result_iterator, 0);` along the lines of:
> file identification using iterator is not supported for directories

christiannicola: Done, thanks for the review!

test/rhash/test_hash_rom.c
christiannicola: Awesome, thanks for the clarification and help.
I believe the change is complete then, all tests are passing on my machine.

Jamiras (approved):

PR #535, opened 2026-08-05
Add RC_CONSOLE_PLAYSTATION_3 (ID 82)

Alasonga:
Adds RC_CONSOLE_PLAYSTATION_3 = 82 to the console enum, as discussed with
@wescopeland in RPCS3/rpcs3#19143. The ID 82 was confirmed by the RA team
as the value to allocate for PS3.

Changes:
- include/rc_consoles.h: add RC_CONSOLE_PLAYSTATION_3 = 82
- src/rcheevos/consoleinfo.c: add "PlayStation 3" to rc_console_name()
- test/rcheevos/test_consoleinfo.c: add test_name and test_memory entries

No memory map is included, that will follow in a separate PR.

PR #536, opened 2026-08-05
rhash: add PS3 hashing support via rc_hash_ps3

Alasonga:
Implements rc_hash_ps3 for PlayStation 3 game identification, following the same pattern as rc_hash_psp. Depends on #535.

PS3 games are identified by hashing two files from the disc image:

PS3_GAME/PARAM.SFO (serial number, title, version)
PS3_GAME/USRDIR/EBOOT.BIN (primary executable)

Reference: https://www.psdevwiki.com/ps3/PARAM.SFO

Verified against real decrypted ISOs in RPCS3:

Demon's Souls (BLUS30443): 7747aaacf594a7dd8b4da33d28e76856
Dark Souls (BLUS30782): c1179907f4af109ab26ba0017888f66b
NieR (BLUS30481): fa0b0fa437e145187faf2a33a242fde2

Changes:

src/rhash/hash_disc.c: add rc_hash_ps3
src/rhash/rc_hash_internal.h: forward declaration
src/rhash/hash.c: add RC_CONSOLE_PLAYSTATION_3 to dispatch and iterators (.iso, .chd)
test/rhash/test_hash_disc.c: add test_hash_ps3, test_hash_ps3_no_param_sfo, test_hash_ps3_no_eboot, test_hash_ps3_m3u


Alasonga:
Note: AI tools were used to assist in development. All code was reviewed, tested, and verified manually before submission.

test/rhash/test_hash_disc.c
```diff
@@ -1434,6 +1474,9 @@ void test_hash_disc(void) {
   TEST(test_hash_psp_video);
   TEST(test_hash_psp_homebrew);
 
+  /* PlayStation 3 */
+  TEST(test_hash_ps3);
+
```
wescopeland, line 1718: Testing: One happy path test case is likely too thin for this diff.

`rc_hash_ps3` has two early returns and neither one is currently exercised by a test case:
* `PS3_GAME/PARAM.SFO` is missing, which returns "Not a PS3 game disc".
* `PARAM.SFO` is found but `PS3_GAME/USRDIR/EBOOT.BIN` is missing, which returns "Could not find primary executable".

Every peer in this file has a negative test, see: `test_hash_psx_cd_no_system_cnf`, `test_hash_sega_cd_invalid_header`, and `test_hash_pce_cd_invalid_header`. The closest match for the second case above is `test_hash_3do_no_launchme`.

_hash.c_ also gained an m3u branch for PS3 and nothing calls it. `test_hash_m3u` already exists as a helper and C64, MSX, and PC-8800 use it via `TEST_PARAMS4`.

src/rhash/hash_disc.c
```diff
@@ -1068,6 +1068,48 @@ int rc_hash_psp(char hash[33], const rc_hash_iterator_t* iterator)
   return rc_hash_finalize(iterator, &md5, hash);
 }
 
+int rc_hash_ps3(char hash[33], const rc_hash_iterator_t* iterator)
```
wescopeland: `rc_hash_ps3` is probably worth forming as a dispatcher, even if there's only one handler today. `rc_hash_wii` in this same file already does this - it reads a magic word, then calls `rc_hash_wii_disc` or `rc_hash_wiiware`.

PS3 has more than one file layout, and RPCS3 supports booting all of them:
* Disc ISO. This uses `PS3_GAME/USRDIR/EBOOT.BIN` and is covered by this PR.
* Extracted disc folder. Uses the same `PS3_GAME` layout on disk.
* HDD game (an installed PSN title). Uses `dev_hdd0/game/{TITLE_ID}/USRDIR/EBOOT.BIN` and has no `PS3_GAME` level.

RPCS3 puts both layouts in one boot search list. See [Emu/System.cpp](https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/System.cpp), which pushes `hdd0_game + "/" + m_title_id + "/USRDIR/"` and `title_path + "/PS3_GAME/USRDIR/"`.

ISO support came from https://github.com/RPCS3/rpcs3/pull/17855. RPCS3 also boots ELF files outside `PS3_GAME/USRDIR/`, from https://github.com/RPCS3/rpcs3/pull/11015.

It's probably okay if folder formats are out of scope for this PR. WiiWare hashing also landed in [a follow-up PR](https://github.com/RetroAchievements/rcheevos/pull/416). Also, the folder formats are both directories, and rhash only opens files. Someone will need to decide which path a client sends.

One thing worth noting for the disc handler: we generally want one game to produce one hash across containers. 3DS handles this in a deliberate fashion, see the "Clear out crypto flags to ensure we get the same hash for decrypted and encrypted ROMs" comment in _hash_encrypted.c_. An extracted disc folder should hash the same as that disc's ISO. Hashing `PARAM.SFO` and `EBOOT.BIN` contents appears to be container-agnostic, which is right, but it's worth being deliberate about since this gets resistant to change once hash entries are registered in our database.

src/rhash/hash_disc.c
```diff
@@ -1068,6 +1068,48 @@ int rc_hash_psp(char hash[33], const rc_hash_iterator_t* iterator)
...
+
+  track_handle = rc_cd_open_track(iterator, 1);
+  if (!track_handle)
+    return rc_hash_iterator_error(iterator, "Could not open track");
+
+  /* https://www.psdevwiki.com/ps3/PARAM.SFO
+   * PS3_GAME/PARAM.SFO contains key/value pairs identifying the game for the system (i.e. serial number,
+   * name, version). PS3_GAME/USRDIR/EBOOT.BIN is the encrypted primary executable.
+   */
+  sector = rc_cd_find_file_sector(iterator, track_handle, "PS3_GAME\\PARAM.SFO", &size);
```
wescopeland, line 1080: Does this read real PS3 images? Two things make me unsure.

`rc_cd_find_file_sector` only handles ISO9660 dir records, and I believe PS3 discs are BD-ROMs. The test builds its image with `generate_iso9660_bin`, so it will pass whether or not real images actually parse.

Redump's canonical PS3 format is the encrypted `.iso`, so I believe a lot of disc dumps in the wild won't parse here regardless. RPCS3 reads those by taking a disc key.

See how RPCS3 handles this, as documented by [the header comment in Loader/ISO.h](https://github.com/RPCS3/rpcs3/blob/f3f52feddfaf4f978b5a519ea1f97cc1a90a479a/rpcs3/Loader/ISO.h). Two details stand out:
1. It decrypts inside `iso_file::read`.
2. The key comes from a `.dkey` or `.key` file next to the `.iso`, or from a 3k3y watermatk at offset 0xF70.

We have a pattern for client-supplied keys in the `encryption` group of `rc_hash_callbacks_t`, added for 3DS.

Which formats have you tested against?

Alasonga:
@wescopeland Before refactoring rc_hash_ps3 into a dispatcher, I want to make sure the architecture covers all formats RPCS3 supports:

Decrypted ISO (ISO 9660): covered by the current implementation
Redump encrypted ISO + .dkey: RPCS3 decrypts sector-by-sector via AES-CBC. Would the right approach be a callback similar to rc_hash_init_3ds_get_ncch_normal_keys_func, where the emulator provides the disc key and rcheevos handles decryption internally?
3k3y encrypted ISO: key is derived from sector data. Same question applies.
Extracted folder: PS3_GAME/USRDIR/EBOOT.BIN directly on the host filesystem. In scope for this PR or a follow-up?

My plan is to refactor rc_hash_ps3 into a dispatcher (like rc_hash_wii) with rc_hash_ps3_disc as the first sub-handler, leaving hooks for the other formats. Should I?

wescopeland:
Hi @Alasonga 👋

One thing from my earlier comment I don't think got covered: which formats have you tested against? "Covered by the current implementation" and "verified against a real decrypted dump" are separate claims. The test only builds a synthetic image via `generate_iso9660_bin`. If you've run this against a real decrypted ISO and it produced a hash, that's worth putting in the PR description.

For the `.dkey` callback, it's worth noting the 3DS case and this one slightly differ. For 3DS, the client holds the keys, so a callback is the only option. Here, the key is a file sitting next to the `.iso`, so rcheevos could read it through its own file callbacks instead. That'd let standalone hashing tools work without emulator support.

On the dispatcher and what belongs in this PR, that's ultimately @Jamiras's call as the code owner and not mine. I'm here to do a first pass and help keep things moving. The dispatcher for `rc_hash_wii` wasn't built ahead of `rc_hash_wiiware` - it arrived later in #416. That likely justifies deferring it (and thus I've marked that comment as resolved).

Alasonga:
@wescopeland updated the description with the tested hashes. Tested against decrypted ISOs only and all produced stable hashes. Encrypted format support via rcheevos file callbacks is being considered; awaiting direction on whether that belongs in this PR or a follow-up.

src/rhash/hash_disc.c
Jamiras: From what I understand, the table of contents is not encrypted, but is stored in UDF instead of ISO-9660.

As such, `rc_cd_find_file_sector` won't work.

Assuming a UDF capable function is created to locate the sectors for `PARAM.SFO` and `EBOOT.BIN`, the read sector callback would be responsible for decrypting the data and we wouldn't need any decryption logic in the hashing code.

Even if the table of contents were encrypted, we could still rely on the read sector callback to do the decryption.

I'm not sure how hard it would be to implement a `rc_udf_find_file_sector` function, but I don't think this PR would be complete without it. While this code won't be used by emulator itself, it would be better to leave the functionality as unsupported than explicitly non-functional.

Jamiras:
Getting caught up on the discussions here.

I assume the hashing code can rely on the sector reading callback to do the decryption and then we wouldn't have to have any explicit additional decryption callbacks.

My only concern is whether the ISO-9660 table of contents reading code (`rc_cd_file_file_sector`) works with UDF table of contents. It sounds like the act of creating the ISO converts the table of contents to ISO-9660, in which case we have to decide whether we want to support encrypted files at all. Will an encrypted disc still have an ISO-9660 table of contents? Or am I totally wrong in assuming the PS3 BD-ROMs have a UDF table of contents? Or maybe they have both?

src/rhash/hash_disc.c
wescopeland: RPCS3's ISO path [uses ISO-9660 rather than UDF](https://github.com/RPCS3/rpcs3/blob/f3f52feddfaf4f978b5a519ea1f97cc1a90a479a/rpcs3/Loader/ISO.cpp#L993-L1041).

It [requires](https://github.com/RPCS3/rpcs3/blob/f3f52feddfaf4f978b5a519ea1f97cc1a90a479a/rpcs3/Loader/ISO.cpp#L57-L68) CD001 at sector 16 before initializing decryption.

`rc_cd_find_file_sector` can locate the files, and the read sector callback should be able to do the decryption.

wescopeland:
@Alasonga Would you mind rebasing onto develop to resolve the merge conflict? It may also be worth chiming in on this discussion: https://github.com/RetroAchievements/rcheevos/pull/536#discussion_r3720207328

Alasonga:
@wescopeland Sorry for the delayed response, I completely lost track of this thread.

@Jamiras The physical disc uses UDF. The act of decrypting and dumping converts it to ISO 9660, CD001 at sector 16, which is what rc_cd_find_file_sector reads. Tested on Dark Souls BLUS30782 and Demon's Souls BLUS30443: both PS3_GAME/PARAM.SFO and PS3_GAME/USRDIR/EBOOT.BIN were found correctly and hashes matched. I didn't have an encrypted ISO to test.

Quick question: would you mind WIP/draft PRs, or do you prefer things more fleshed out before opening? I've made significantly more progress on the PS3 support side, even though I'm constantly revisiting and refining things, but I've been holding off on posting because I wasn't sure of your preference.

wescopeland:
> Quick question: would you mind WIP/draft PRs, or do you prefer things more fleshed out before opening? I've made significantly more progress on the PS3 support side, even though I'm constantly revisiting and refining things, but I've been holding off on posting because I wasn't sure of your preference.

Probably a Jamiras question. I don't have a strong opinion one way or another.

Jamiras:
> Quick question: would you mind WIP/draft PRs, or do you prefer things more fleshed out before opening? I've made significantly more progress on the PS3 support side, even though I'm constantly revisiting and refining things, but I've been holding off on posting because I wasn't sure of your preference.

I generally prefer not to look at something until it's ready to review, unless there's something specific you want someone to look at before the entire PR is ready. In those cases, it's probably easiest to just send me a link to the code in your branch without opening a PR.

Jamiras (approved):
Okay, so after I figured out how to decrypt my dumps, I get the same hashes for Nier and Dark Souls, so I think this is probably sufficient.

It may still be worth persuing adding the decryption support as the encrypted dumps did not generate the same hashes.

> Redump encrypted ISO + .dkey: RPCS3 decrypts sector-by-sector via AES-CBC. Would the right approach be a callback similar to rc_hash_init_3ds_get_ncch_normal_keys_func, where the emulator provides the disc key and rcheevos handles decryption internally?

> For the .dkey callback, it's worth noting the 3DS case and this one slightly differ. For 3DS, the client holds the keys, so a callback is the only option. Here, the key is a file sitting next to the .iso, so rcheevos could read it through its own file callbacks instead. That'd let standalone hashing tools work without emulator support.

But that could be a separate PR, and I don't think it necessarily has to be part of rcheevos as long as the decrypted path serves as a reference implementation.

PR #539, opened 2026-08-22
Fix out-of-bounds string comparisons in libretro denylist

stenzek:
memcmp() would read out-of-bounds when the incoming library name was shorter than the table name.

Use length-safe string comparisons for libretro core, setting, and value lookup. Add short and long input regressions that previously triggered global-buffer-overflow under ASan.


test/test_rc_libretro.c
Jamiras, line 1: I feel like these tests should all be grouped together with a comment indicating they're validating boundary reads. Without any context as to what they're doing, they seem unnecessary where they're at.

stenzek: Done.

Jamiras (approved):

PR #541, opened 2026-08-25
Fix numerous bounds-related issues

stenzek:
See commit descriptions for details.

There's a few more I still need to look at, but as these were agent findings I'm going through them with a fine-toothed comb before upstreaming. Already had to make a few corrections on these alone...

### DIFF
--- a/src/rapi/rc_api_common.c
+++ b/src/rapi/rc_api_common.c
@@ -711,9 +711,12 @@ int rc_json_get_string(const char** out, rc_buffer_t* buffer, const rc_json_fiel
 
     *out = dst = (char*)rc_buffer_reserve(buffer, len - 1); /* -2 for quotes, +1 for null terminator */
 
-    do {
+    while (src < field->value_end && *src != '\"') {
       if (*src == '\\') {
         ++src;
+        if (src >= field->value_end)
+          return 0;
+
         if (*src == 'n') {
           /* newline */
           ++src;
@@ -731,7 +734,7 @@ int rc_json_get_string(const char** out, rc_buffer_t* buffer, const rc_json_fiel
         if (*src == 'u') {
           /* unicode character */
           uint32_t ucs32_char;
-          if (src + 5 >= field->value_end) /* incomplete unicode character */
+          if ((size_t)(field->value_end - src) < 6) /* incomplete unicode character */
             return 0;
 
           ucs32_char = rc_json_decode_hex4(src + 1);
@@ -739,7 +742,7 @@ int rc_json_get_string(const char** out, rc_buffer_t* buffer, const rc_json_fiel
 
           if (ucs32_char >= 0xD800 && ucs32_char < 0xE000) {
             /* surrogate lead - look for surrogate tail */
-            if (ucs32_char < 0xDC00 && src[0] == '\\' && src[1] == 'u') {
+            if (ucs32_char < 0xDC00 && (size_t)(field->value_end - src) >= 6 && src[0] == '\\' && src[1] == 'u') {
               const uint32_t surrogate = rc_json_decode_hex4(src + 2);
               src += 6;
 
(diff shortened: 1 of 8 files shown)


src/rapi/rc_api_common.c
```diff
@@ -731,16 +736,22 @@ int rc_json_get_string(const char** out, rc_buffer_t* buffer, const rc_json_fiel
...
           src += 5;
 
           if (ucs32_char >= 0xD800 && ucs32_char < 0xE000) {
             /* surrogate lead - look for surrogate tail */
-            if (ucs32_char < 0xDC00 && src[0] == '\\' && src[1] == 'u') {
-              const uint32_t surrogate = rc_json_decode_hex4(src + 2);
+            if (ucs32_char < 0xDC00 && (size_t)(field->value_end - src) >= 2 && src[0] == '\\' && src[1] == 'u') {
+              uint32_t surrogate;
+
+              /* the six-byte tail must be followed by the string's closing quote */
```
Jamiras: No it doesn't. It can be followed by more characters before the closing quote. The lack of a closing quote should be limited to the loop check.

src/rapi/rc_api_common.c
```diff
@@ -731,16 +736,22 @@ int rc_json_get_string(const char** out, rc_buffer_t* buffer, const rc_json_fiel
...
             return 0;
 
           ucs32_char = rc_json_decode_hex4(src + 1);
           src += 5;
 
           if (ucs32_char >= 0xD800 && ucs32_char < 0xE000) {
             /* surrogate lead - look for surrogate tail */
-            if (ucs32_char < 0xDC00 && src[0] == '\\' && src[1] == 'u') {
-              const uint32_t surrogate = rc_json_decode_hex4(src + 2);
+            if (ucs32_char < 0xDC00 && (size_t)(field->value_end - src) >= 2 && src[0] == '\\' && src[1] == 'u') {
```
Jamiras: This should probably check for size >= 6 to eliminate the additional check down a couple lines.

If size < 6, it will result in a placeholder character and the next iteration of the loop will find the incomplete character and bail then.

src/rc_util.c
```diff
@@ -65,10 +65,15 @@ uint8_t* rc_buffer_reserve(rc_buffer_t* buffer, size_t amount)
...
+       * to the chunk header, and the remaining will be used for data. avoid integer overflow by bounding
+       * the allocation size.
        */
       const size_t chunk_header_size = sizeof(rc_buffer_chunk_t);
-      const size_t alloc_size = (chunk_header_size + amount + 0xFF) & ~0xFF;
+      size_t alloc_size;
+      if (amount > (chunk_header_size + 0xFF) && amount > (SIZE_MAX - chunk_header_size - 0xFF))
+        break;
+
+      alloc_size = (chunk_header_size + amount + 0xFF) & ~0xFF;
```
Jamiras: If we're trying to allocate a chunk large enough to trigger this code, we have other problems.

I think I'd rather have an early abort above the while statement.
```
if (amount >= 0x40000000) /* we should never need 1GB of space for anything */
    return NULL;

while (chunk)
```

src/rapi/rc_api_common.c
stenzek: Agreed! Done.

src/rc_util.c
stenzek: Yeah, I was a bit torn on this one. But this is a good suggestion, I was also just considering an assertion for debug builds, because we really shouldn't be hitting this in the first place.

A related note, `rc_buffer_strncpy()` and friends could technically overflow too on the +1, but the callers were all using lengths derived from incoming data, which couldn't possibly be `SIZE_MAX`. So I left those alone rather than muddying things up, but tools (and agents) will probably still flag it.

Jamiras (approved):

PR #542, opened 2026-08-27
allow {recall} of const to match across alts

Jamiras:
Fixes an issue where the parse would overflow the buffer when using "Remember {constant}" in multiple alts.

https://discord.com/channels/310192285306454017/310195377993416714/1542379959941734461

Because the preparse reuses the same data structures for analysis, the comparison that "Remember {constant}" equals "Remember {constant}" was true, but the regular parse thought they different instances and therefore didn't share them, resulting in a second allocation that was not accounted for when calculating the buffer size. Then, anything chaining off the new root node also resulted in an extra allocation as the parents differed.

Given the amount of stuff chaining off the first "Remeber {constant}" in the affected achievement, the write exceeded the buffer size significantly, leading to the crash.

### DIFF
--- a/src/rcheevos/condset.c
+++ b/src/rcheevos/condset.c
@@ -363,7 +363,8 @@ rc_condset_t* rc_parse_condset(const char** memaddr, rc_parse_state_t* parse) {
 
         classification = combining_classification;
       }
-      else {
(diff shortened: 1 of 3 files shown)


suspect15:
Much thanks for the incredibly fast problem analysis and fix.

PR #545, opened 2026-09-03
rcheevos: add PS3 memory map to consoleinfo.c

Alasonga:
Adds rc_memory_regions_playstation3 to consoleinfo.c. Depends on #535

## Memory regions

| RA address range | Size | Type | Description |
|---|---|---|---|
| 0x00000000–0x0FFFFFFF | 256 MB | SYSTEM_RAM | Main RAM (preallocated) |
| 0x10000000–0x2FFFFFFF | 512 MB | SYSTEM_RAM | RSX Context + User RAM (64K pages, lazy) |
| 0x30000000–0x3FFFFFFF | 256 MB | SYSTEM_RAM | User RAM (64K pages, lazy) |
| 0x40000000–0x4FFFFFFF | 256 MB | SYSTEM_RAM | User RAM (1M pages, lazy) |

RSX context (0x10000000–0x1FFFFFFF) is GPU DMA memory allocated by sys_rsx_context_allocate before any sys_memory_allocate call. It is fused with the first half of the user64K pool into a single region since they are physically contiguous and game state is not expected in RSX context space.

## Address mapping

RA addresses are a 1:1 map to RPCS3 virtual addresses. The RPCS3 read_memory callback returns zeroed bytes for lazy regions that have not yet been allocated, preventing rcheevos from permanently disabling achievements that reference unallocated memory at game load time.

## References

- `_find_map` scan from 0x10000000: https://github.com/RPCS3/rpcs3/blob/a0bebfa5c/rpcs3/Emu/Memory/vm.cpp#L1893
- RSX context 256 MB reservation: https://github.com/RPCS3/rpcs3/blob/a0bebfa5c/rpcs3/Emu/Cell/lv2/sys_rsx.cpp#L272
- user64K 512 MB hardcoded pool: https://github.com/RPCS3/rpcs3/blob/a0bebfa5c/rpcs3/Emu/Cell/lv2/sys_memory.cpp#L105

## Tests

All tests pass (`make HAVE_HASH=1` in `test/`).


Alasonga:
Note: AI tools were used to assist in development. All code was reviewed, tested, and verified manually before submission.

src/rcheevos/consoleinfo.c
```diff
@@ -830,6 +830,16 @@ static const rc_memory_region_t _rc_memory_regions_psp[] = {
 };
 static const rc_memory_regions_t rc_memory_regions_psp = { _rc_memory_regions_psp, 3 };
 
+/* ===== PlayStation 3 ===== */
+/* https://github.com/RPCS3/rpcs3/blob/master/rpcs3/Emu/Memory/vm.cpp */
```
Jamiras: Could you narrow down where in this file the below mapping was derived from?

The only thing I really see is [this](https://github.com/RPCS3/rpcs3/blob/33a723affa1554ef310a138d3c8810b29c3aa43c/rpcs3/Emu/Memory/vm.cpp#L2230-L2239), and it doesn't clearly delineate where each block starts/stops.

Jamiras (commented):
I don't have sufficient understanding to validate the map itself, but the implementation is approved.

src/rcheevos/consoleinfo.c
Alasonga: The g_locations block you found is the right area, but the actual addresses come from three separate files.

_find_map in vm.cpp scans upward from 0x10000000 and assigns that base to the first caller. sys_rsx_context_allocate (triggered by cellGcmInit) always runs before any sys_memory_allocate, so RSX context wins the first slot at 0x10000000 with a hardcoded 256 MB size. The user64K pool then follows at 0x20000000 with its size hardcoded to 512 MB, regardless of the game's actual request. User1M lands at 0x40000000.

RSX context (0x10000000–0x1FFFFFFF) is GPU DMA memory, game state is not expected there based on RPCS3 source analysis, though not exhaustively verified across all titles. It is fused with the first half of user64K into a single region for that reason.

The PR will be updated with corrected addresses and these references replacing the file-level link.

References
1. _find_map, scan from 0x10000000: https://github.com/RPCS3/rpcs3/blob/a0bebfa5c/rpcs3/Emu/Memory/vm.cpp#L1893
2. RSX context (256 MB) reservation: https://github.com/RPCS3/rpcs3/blob/a0bebfa5c/rpcs3/Emu/Cell/lv2/sys_rsx.cpp#L272
3. user64K (512 MB) hardcoded pool: https://github.com/RPCS3/rpcs3/blob/a0bebfa5c/rpcs3/Emu/Cell/lv2/sys_memory.cpp#L105

Sorry about the misunderstanding. It's now set correctly in the code.

Jamiras (approved):
I'm still not confident that the map is correct, but have to assume it is as the author is more fluent with the target architecture.

src/rcheevos/consoleinfo.c
Jamiras: I can see how the first maps to $10000000, but I don't see how the other two map to any specific address. The comment still doesn't exactly explain the maps, but at this point, I'm not sure that it can.

Jamiras:
> The RPCS3 read_memory callback returns zeroed bytes for lazy regions that have not yet been allocated, preventing rcheevos from permanently disabling achievements that reference unallocated memory at game load time.

That's an accurate, though unfortunate observation. https://github.com/RetroAchievements/RAIntegration/pull/1314 excludes blocks from the search results if the read function returns 0 bytes read. But rc_client will also disable achievements if the read returns 0 bytes and it's not part of a pointer chain.https://github.com/RetroAchievements/rcheevos/blob/e871ef53ea87c4b4a7643957e8f6084407ffe837/src/rc_client.c#L5884-L5896

Since the memory isn't pre-distributed we can't just switch read functions after the game loads (use one before load to return dummy data saying the addresses are valid and one after load returning no data for unallocated regions).

As a result, the search space will immediately be 1.3GB (256MBx5), which is the problem the above PR was trying to address. I wonder if we should allow `peek_memory` to return 0xFFFFFFFF to indicate the address is valid but not yet available?

Alasonga:
> > The RPCS3 read_memory callback returns zeroed bytes for lazy regions that have not yet been allocated, preventing rcheevos from permanently disabling achievements that reference unallocated memory at game load time.
> 
> That's an accurate, though unfortunate observation. [RetroAchievements/RAIntegration#1314](https://github.com/RetroAchievements/RAIntegration/pull/1314) excludes blocks from the search results if the read function returns 0 bytes read. But rc_client will also disable achievements if the read returns 0 bytes and it's not part of a pointer chain.
> 
> https://github.com/RetroAchievements/rcheevos/blob/e871ef53ea87c4b4a7643957e8f6084407ffe837/src/rc_client.c#L5884-L5896
> 
> Since the memory isn't pre-distributed we can't just switch read functions after the game loads (use one before load to return dummy data saying the addresses are valid and one after load returning no data for unallocated regions).
> 
> As a result, the search space will immediately be 1.3GB (256MBx5), which is the problem the above PR was trying to address. I wonder if we should allow `peek_memory` to return 0xFFFFFFFF to indicate the address is valid but not yet available?

There are actually two separate read callbacks in the RPCS3 integration: read_memory for rc_client and readBlock for RAIntegration. So the two concerns can be handled independently.

read_memory already returns zeroed bytes with num_bytes for lazy unallocated regions, so rc_client sees a valid read and achievements aren't invalidated. No sentinel needed.

 // vm::try_access reads PS3 virtual memory; returns false for uncommitted lazy pages
  static u32 read_memory(u32 address, u8* buffer, u32 num_bytes, rc_client_t*)
   {
        if (vm::try_access(address, buffer, num_bytes, false))
            return num_bytes;
                 
        // lazy banks: return zeros so rc_client sees a valid read, not a failure
        if ((address >= s_bank1_base && address < s_bank1_base + s_bank1_size) ||
            (address >= s_bank2_base && address < s_bank2_base + s_bank2_size) ||
            (address >= s_bank3_base && address < s_bank3_base + s_bank3_size))
        {
            memset(buffer, 0, num_bytes);
            return num_bytes;
        }
        return 0;
   }
                                                                               
For the search space, the block reader currently zero-fills unconditionally, which bypasses #1314. Fixing it to return 0 for fully unallocated blocks will let #1314 handle that side naturally.

Jamiras:
> There are actually two separate read callbacks in the RPCS3 integration: read_memory for rc_client and readBlock for RAIntegration. So the two concerns can be handled independently.

We shouldn't rely on this behavior. When the DLL is loaded through rc_client via [rc_client_begin_load_raintegration](https://github.com/RetroAchievements/rcheevos/wiki/rc_client_begin_load_raintegration), generic `ReadBlock` handlers are registered that wrap the `read_memory` callback.

In any case, it's something we don't have to address immediately.

Alasonga:
Following up on the search space discussion: opened RAIntegration PR #1348 which adds `_RA_InstallSearchMemoryBankReader` — a dedicated per-bank block reader for `CaptureMemory`. It's registered independently of `_RA_InstallMemoryBankBlockReader` and the underlying read callback, so it works correctly regardless of how `readBlock` is wired internally. Validated across 6 PS3 games: search space drops from 1280 MB to ~430 MB on average.

PR #546, opened 2026-09-05
Support the current libctru 3DS define

matbo87:
Add `__3DS__` to the RecursiveLock platform check, while keeping `_3DS` for compatibility.


Jamiras (approved):
Would have prefered a link to the documentation explaining this change, but I found one: https://github.com/devkitPro/libctru/blob/master/Changelog.md#version-210

